PR #12, opened 2018-10-08
Initial module setup

natali-zaitseva:
## **Purpose**
Per 
https://issues.folio.org/browse/MODKBEKBJ-14 
https://issues.folio.org/browse/MODKBEKBJ-45
https://issues.folio.org/browse/MODKBEKBJ-44
https://issues.folio.org/browse/MODKBEKBJ-16
https://issues.folio.org/browse/MODKBEKBJ-15
https://issues.folio.org/browse/MODKBEKBJ-3
we are supposed to create an initial file set for the module 
## Approach
* Generate pom and all related files using https://github.com/folio-org/mod-rmb-template 
* Create ModuleDescriptor and DeploymentDescriptor files
* Add raml samples files
* Copy raml examples from the https://github.com/folio-org/mod-kb-ebsco/tree/master/ramls/examples

PR #13, opened 2018-10-10
MODKBEKBJ-40: GET /eholdings/configuration endpoint

andrii-paias:
## Purpose
Implement GET /eholdings/configuration endpoint

## Approach
Used ruby mod-kb-ebsco as reference.
Adapted RMAPIConfiguration class from mod-codex-ekb module to get configuration from RMAPI.
Used Wiremock to implement tests simlar to tests from ruby module.

#### TODOS and Open Questions
* We need to wait until [initial module setup PR](https://github.com/folio-org/mod-kb-ebsco-java/pull/12 ) is merged, before merging this PR. 
This PR is only for commit https://github.com/folio-org/mod-kb-ebsco-java/pull/13/commits/a5d661fd279d27b474bc5733f67366481bf021b8

* Ruby module returns several field that seem to be specific to jsonapi:
Response from ruby module
```json
{
    "data": {
        "id": "configuration",
        "type": "configurations",
        "attributes": {
            "customerId": "apidvcorp",
            "apiKey": "****************************************",
            "rmapiBaseUrl": "https://api.ebsco.io"
        }
    },
    "jsonapi": {
        "version": "1.0"
    }
}
```
Current response from java implementation
```json
{
    "data": {
        "type": "configurations",
        "attributes": {
            "customerId": "apidvcorp",
            "apiKey": "****************************************",
            "rmapiBaseUrl": "https://api.ebsco.io"
        }
    }
}
```
Ruby module has "id": "configuration", and   "jsonapi": { "version": "1.0" }.
Those fields are not part of RAML definition, so I am not sure if we need to return them from java module.

### DIFF
--- a/src/main/java/org/folio/config/RMAPIConfigurationClient.java
+++ b/src/main/java/org/folio/config/RMAPIConfigurationClient.java
@@ -6,8 +6,6 @@ import org.folio.http.ConfigurationClientProvider;
 import org.folio.rest.client.ConfigurationsClient;
 import org.folio.rest.tools.client.Response;
 import org.folio.rest.tools.utils.TenantTool;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.URL;
@@ -17,8 +15,6 @@ import java.util.concurrent.CompletableFuture;
  * Retrieves the RM API connection details from mod-configuration.
  */
 public class RMAPIConfigurationClient {
-
-  private static final Logger LOG = LoggerFactory.getLogger(RMAPIConfigurationClient.class);
 
   private static final String EBSCO_URL_CODE = "kb.ebsco.url";
   private static final String EBSCO_API_KEY_CODE = "kb.ebsco.apiKey";
@@ -47,19 +43,22 @@ public class RMAPIConfigurationClient {
 
     try {
       URL url = new URL(okapiURL);
-      int port = url.getPort() != -1? url.getPort() : url.getDefaultPort();
+      int port = url.getPort() != -1 ? url.getPort() : url.getDefaultPort();
       ConfigurationsClient configurationsClient = configurationClientProvider.createClient(url.getHost(), port, tenantId, apiToken);
       configurationsClient.getEntries("module=EKB", 0, QUERY_LIMIT, null, null, response ->
         response.bodyHandler(body -> {
-          if (!Response.isSuccess(response.statusCode())) {
-            LOG.error("Cannot get configuration data: error code - {} response body - {}", response.statusCode(), body);
-            future.completeExceptionally(new IllegalStateException(body.toString()));
+          try {
+            if (!Response.isSuccess(response.statusCode())) {
+              throw new IllegalStateException("Cannot get configuration data: error code - " +
+                response.statusCode() + " response body - " + body.toString());
+            }
+            JsonObject entries = body.toJsonObject();
+            future.complete(mapResults(entries.getJsonArray("configs")));
+          } catch (Exception e) {
+            future.completeExceptionally(e);
           }
-          JsonObject entries = body.toJsonObject();
-          future.complete(mapResults(entries.getJsonArray("configs")));
         }));
     } catch (IOException e) {
-      LOG.error("Cannot get configuration data: " + e.getMessage(), e);
       future.completeExceptionally(e);
     }
     return future;
--- a/src/main/java/org/folio/config/RMAPIConfigurationClient.java
+++ b/src/main/java/org/folio/config/RMAPIConfigurationClient.java
@@ -34,7 +34,9 @@ public class RMAPIConfigurationClient {
    * Returns the RM API configuration for the tenant.
    *
    * @param apiToken token used to call okapi
-   * @return The RMI API configuration for the tenant.
+   * @param tenant tenant for which configuration is retrieved
+   * @param okapiURL url of OKAPI server
+   * @return The RM API configuration for the tenant.
    */
   public CompletableFuture<RMAPIConfiguration> retrieveConfiguration(String apiToken, String tenant, String okapiURL) {
     final String tenantId = TenantTool.calculateTenantId(tenant);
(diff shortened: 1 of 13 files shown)


dmtkachenko:
> 
> 
> Looking at UIEH and seeing if jsonapi - version and data.id are being used,, I see a couple of mock responses having jsonapi.version -> which I think can be adjusted on the UI part but data.id is also being used in several places -> so probably, important to get both of them in there by adjusting RAML for this endpoint to generate those objects; schema for which can be referenced from here: https://github.com/sduvvuri-ebsco/POC/tree/test_branch/ramls
> 
> Also, agreed that this PR should go after the initial setup and at that point, hopefully, we'll not see this many files changed.



> 
> 
> Looking at UIEH and seeing if jsonapi - version and data.id are being used,, I see a couple of mock responses having jsonapi.version -> which I think can be adjusted on the UI part but data.id is also being used in several places -> so probably, important to get both of them in there by adjusting RAML for this endpoint to generate those objects; schema for which can be referenced from here: https://github.com/sduvvuri-ebsco/POC/tree/test_branch/ramls
> 
> Also, agreed that this PR should go after the initial setup and at that point, hopefully, we'll not see this many files changed.

Since we agreed to tweak the responses according to jsonapi requirements in a separate US for each of the endpoints, I think we can go ahead with the current implementation w/o any changes.
PS: I create a story for Conf endpoint: https://issues.folio.org/browse/MODKBEKBJ-48

src/main/java/org/folio/rest/converter/RMAPIConfigurationConverter.java
```diff
@@ -0,0 +1,21 @@
+package org.folio.rest.converter;
+
+import org.apache.commons.lang3.StringUtils;
+import org.folio.config.RMAPIConfiguration;
+import org.folio.rest.jaxrs.model.EholdingsConfigurationPutApplicationVndApiJsonImpl;
+
+/**
+ * Converts internal RMAPIConfiguration object into object that will be returned in response
+ */
+public class RMAPIConfigurationConverter {
```
sduvvuri-ebsco: @andrii-paias : What do you think of renaming this class to RMAPIConfigurationSerializer instead of Converter? Or is Converter better understandable in Java space?

src/main/java/org/folio/rest/impl/EholdingsConfigurationImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+    configurationClient.retrieveConfiguration(
+      okapiHeaders.get(HeaderConstants.OKAPI_TOKEN_HEADER),
+      okapiHeaders.get(HeaderConstants.OKAPI_TENANT_HEADER),
+      okapiHeaders.get(HeaderConstants.OKAPI_URL_HEADER))
+      .thenAccept(rmAPIconfiguration -> {
+        EholdingsConfigurationPutApplicationVndApiJsonImpl configuration = converter.convert(rmAPIconfiguration);
+        asyncResultHandler.handle(Future.succeededFuture(GetEholdingsConfigurationResponse.respond200WithApplicationVndApiJson(configuration)));
+      })
+      .exceptionally(throwable -> {
```
sduvvuri-ebsco: Not sure about this, if this will end up throwing a stack trace, instead can we throw an InternalServerError?

src/main/java/org/folio/rest/validator/HeaderValidator.java
```diff
@@ -0,0 +1,42 @@
...
+public class HeaderValidator {
+
+  private final Collection<String> expectedHeaders = Arrays.asList(
+    HeaderConstants.OKAPI_TOKEN_HEADER,
+    HeaderConstants.OKAPI_URL_HEADER,
+    HeaderConstants.OKAPI_TENANT_HEADER);
+
+  /**
+   * @param okapiHeaders request headers
+   * @param asyncResultHandler handler that will be called with error response if headers are invalid invalid
```
sduvvuri-ebsco: a very minor nitpick - 2 invalids in comments

src/main/java/org/folio/rest/impl/EholdingsConfigurationImpl.java
andrii-paias: This returns code 500 with body "Server Error"
now that I looked into it closer I changed it to always log exception before returning error.

src/main/java/org/folio/rest/converter/RMAPIConfigurationConverter.java
andrii-paias: My reasoning was that serializer normally means a class that serializes an object to json string, or has something to do with java Serializable interface, so I named it converter. But I don't have a strong opinion about it, if this will make it consistent with other modules we can rename it to serializer.

src/main/java/org/folio/rest/impl/EholdingsConfigurationImpl.java
andrii-paias: By throwing InternalServerError do you mean returning 500 or something else?

src/main/java/org/folio/rest/converter/RMAPIConfigurationConverter.java
sduvvuri-ebsco: I do not have a strong opinion either. Let's leave it the way it is for now.

src/main/java/org/folio/rest/impl/EholdingsConfigurationImpl.java
sduvvuri-ebsco: I was thinking something on these lines: 
`asyncResultHandler.handle(Future.succeededFuture(Response.respond500WithTextPlain(throwable.getCause().getMessage())));` - where the Response class is whatever class we are using there which in this case is `GetEholdingsConfigurationResponse`

andrii-paias: Sure, will add this to raml:
```      500:
        description: "Internal server error"
        body:
          text/plain:
            example: "Internal server error, contact administrator"
```

also this is a case that catches all exceptions, so throwable.getCause().getMessage() might not have a high level message, if it is not ok, then we can use "Internal server error" for all exceptions and log the message with stacktrace

src/main/java/org/folio/config/RMAPIConfigurationClient.java
```diff
@@ -0,0 +1,97 @@
...
+    this.configurationClientProvider = configurationClientProvider;
+  }
+
+  /**
+   * Returns the RM API configuration for the tenant.
+   *
+   * @param apiToken token used to call okapi
+   * @return The RMI API configuration for the tenant.
+   */
+  public CompletableFuture<RMAPIConfiguration> retrieveConfiguration(String apiToken, String tenant, String okapiURL) {
```
natali-zaitseva: Could you please update java doc with info about other parameters (tenant, okapiURL)

PR #14, opened 2018-10-16
MODKBEKBJ-48: Update Configurations RAML to include needed schema for JSON API

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-48, we are supposed to adjust GET and PUT `/eholdings/configuration` endpoint's RAML to be able to accept/produce requests/responses by conforming to JSON API schema so that they match with what mod-kb-ebsco currently accepts/produces so we do not have to modify UIEH in any way.

## Approach
- Put together JSON schema for data, attributes, errors, json api etc. which is the structure used for requests/responses to and from the `/eholdings/configuration` endpoint.
- Ensure by running raml-cop that RAML is valid
- Ensure that build is successful

#### TODOS and Open Questions
- [ ] After this gets merged, we have to modify https://github.com/folio-org/mod-kb-ebsco-java/pull/13 to use the POJOs that will be generated by this RAML and verify that requests/responses abide by JSON API schema. Tweak/Adjust RAML/JSON schemas put together above based on how the modification to the mentioned PR goes.


natali-zaitseva:
@sduvvuri-ebsco I am just curious about the file jsonapi.json. Do some logic relies on the information placed in this file? I have looked through the json files inside other modules and did not see any of them using or requires info about jsonapi version used. Did I missed something or we can get rid of it?
Thanks in advance.

sduvvuri-ebsco:
@natali-zaitseva : Thanks much for taking a look at the PR. Other modules do not use JSON API: https://jsonapi.org/ where as mod-kb-ebsco module uses json api schema to model its requests and responses around. UI-Eholdings, the front-end, uses that structure for parsing data from mod-kb-ebsco, so, in order to not change anything on the front-end, we decided to stick to using JSON API format for requests and responses. That's the only reason we need it. No functional logic because of it.

PR #19, opened 2018-10-18
MODKBEKBJ-41: PUT /eholdings/configuration endpoint

andrii-paias:
## Purpose
Implement PUT configuration endpoint

## Approach
Verify that new configuration is valid by sending get request to RM API,
then use delete and post requests to mod-configuration to replace existing configuration with new one.

#### TODOS and Open Questions
- [ ] I copied RMAPIService class from mod-codex-ekb to send request to RM API, is there a way to move this class to some common module and then reuse it in both mod-codex-ekb and mod-kb-ebsco-java modules?


src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -0,0 +1,113 @@
...
+    final HttpClientRequest request = httpClient.getAbs(query);
+
+    request.headers().add(HTTP_HEADER_ACCEPT, APPLICATION_JSON);
+    request.headers().add(HTTP_HEADER_CONTENT_TYPE, APPLICATION_JSON);
+    request.headers().add(RMAPI_API_KEY, apiKey);
+
+    LOG.info("RMAPI Service absolute URL is" + request.absoluteURI());
+
+    request.handler(response -> response.bodyHandler(body -> {
+      if (response.statusCode() == 200) {
```
dmtkachenko: I think you can close httpClient in the first statement of handler and don't do this for 2 different if branches
see how similar things done with connection (2nd code block):
https://vertx.io/docs/guide-for-java-devs/#_wiki_verticle_initialization_phases

PR #23, opened 2018-10-22
MODKBEKBJ-39: GET /eholdings/status endpoint

andrii-paias:
## Purpose
Implement GET status endpoint

## Approach
Implemented caching of current configuration and its status.
Reused method method from "PUT configuration" to validate configuration.

#### TODOS and Open Questions
- [x] I hard-coded cache expiration period to two minutes, is there a conventional way to define expiration period in a property file and inject it into cache class?

### DIFF
--- a/src/main/java/org/folio/config/RMAPIConfiguration.java
+++ b/src/main/java/org/folio/config/RMAPIConfiguration.java
@@ -7,6 +7,7 @@ public final class RMAPIConfiguration {
   private String customerId;
   private String apiKey;
   private String url;
+  private Boolean isValid;
 
   public String getCustomerId() {
     return customerId;
@@ -20,6 +21,10 @@ public final class RMAPIConfiguration {
     return url;
   }
 
+  public Boolean getValid() {
+    return isValid;
+  }
+
   public void setCustomerId(String customerId) {
     this.customerId = customerId;
   }
@@ -30,6 +35,10 @@ public final class RMAPIConfiguration {
 
   public void setUrl(String url) {
     this.url = url;
+  }
+
+  public void setValid(Boolean valid) {
+    isValid = valid;
   }
 
   @Override
--- a/src/main/java/org/folio/config/RMAPIConfiguration.java
+++ b/src/main/java/org/folio/config/RMAPIConfiguration.java
@@ -7,7 +7,7 @@ public final class RMAPIConfiguration {
   private String customerId;
   private String apiKey;
   private String url;
-  private Boolean isValid;
+  private Boolean configValid;
 
   public String getCustomerId() {
     return customerId;
@@ -21,8 +21,8 @@ public final class RMAPIConfiguration {
     return url;
   }
 
-  public Boolean getValid() {
-    return isValid;
+  public Boolean getConfigValid() {
+    return configValid;
   }
 
   public void setCustomerId(String customerId) {
@@ -37,8 +37,8 @@ public final class RMAPIConfiguration {
     this.url = url;
   }
 
-  public void setValid(Boolean valid) {
-    isValid = valid;
+  public void setConfigValid(Boolean configValid) {
+    this.configValid = configValid;
   }
 
   @Override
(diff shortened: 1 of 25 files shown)


dmtkachenko:
> I hard-coded cache expiration period to two minutes, is there a conventional way to define expiration period in a property file and inject it into cache class?

I agree it has to be configurable. From what I've seen till the moment configuration can be defined
1. in verticle settings and accessed by config() method
2. using old-fashion Property file. you have to read the properties manually

I'll try to find a better way and get back to you

src/main/java/org/folio/config/cache/RMAPIConfigurationCache.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import org.folio.config.RMAPIConfiguration;
+import org.folio.properties.PropertyConfiguration;
+
+import java.util.concurrent.TimeUnit;
+
+public final class RMAPIConfigurationCache {
+  private static final String CACHE_KEY = "Configuration";
```
carolegodfrey: As mentioned at standup -- RM API Configuration is tenant specific. So CACHE_KEY will need to either need to be or include the `x-okapi-tenant`

src/main/java/org/folio/config/cache/RMAPIConfigurationCache.java
```diff
@@ -0,0 +1,37 @@
...
+  private final Cache<String, RMAPIConfiguration> cache;
```
carolegodfrey: As mentioned at standup -- RM API Configuration is tenant specific. So CACHE_KEY will need to either be or include the `x-okapi-tenant`

src/main/java/org/folio/config/RMAPIConfiguration.java
```diff
@@ -7,6 +7,7 @@
   private String customerId;
   private String apiKey;
   private String url;
+  private Boolean isValid;
```
sduvvuri-ebsco: can we please rename this variable to be "getConfigurationValid" instead of "isValid" for more clarity?

src/main/java/org/folio/config/RMAPIConfiguration.java
```diff
@@ -20,6 +21,10 @@ public String getUrl() {
     return url;
   }
 
+  public Boolean getValid() {
```
sduvvuri-ebsco: Possibly, similar renaming with "getConfigValid" and "set**"?

src/main/java/org/folio/config/RMAPIConfigurationServiceCache.java
```diff
@@ -0,0 +1,55 @@
...
+
+  public RMAPIConfigurationServiceCache(RMAPIConfigurationService rmapiConfigurationService) {
+    this.rmapiConfigurationService = rmapiConfigurationService;
+  }
+
+  @Override
+  public CompletableFuture<RMAPIConfiguration> retrieveConfiguration(OkapiData okapiData) {
+    RMAPIConfiguration cachedConfiguration = RMAPIConfigurationCache.getInstance()
+      .getValue();
+    if(cachedConfiguration != null){
```
sduvvuri-ebsco: one thing that we have to be careful about is if there is a possibility for stale data in cache? I see that if the user updates configuration making a PUT request to mod-kb-ebsco-java, then, yes we are updating cache with latest config data. But there is a possibility of other scenarios like making POST requests directly to configuration module - in case of which our cached data might be stale.

andrii-paias: It is possible for configuration to become stale if someone sends POST requests to mod-configuration, but cache will expire after several minutes and will be updated.

src/main/java/org/folio/config/cache/RMAPIConfigurationCache.java
andrii-paias: Thanks, 
updated PR to use tenant as cache key

src/main/java/org/folio/config/RMAPIConfigurationServiceCache.java
sduvvuri-ebsco: yep, that's true. Hopefully, if we are using stale data too, verification will come back false and we should then be able to make an actual request to get latest!

PR #24, opened 2018-10-26
MODKBEKBJ-34 - Adjust RAML for Titles endpoint to conform to JSON API

natali-zaitseva:
* updated raml files for title endpoints

## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-34,  we want to tweak RAML and put together JSON schema that supports requests/responses for various /eholdings/titles endpoints conforming to JSON API schema. 

## Approach
* organize raml directory
* tweak json schema for the following endpoints:
  * GET and POST for /eholdings/titles
  * GET /eholdings/titles/{titleId}

## TODOS and Open Questions

The types 'visibilityData' and 'proxy' caused Jenkins errors because they were created not as classes but interfaces by RMB which is not acceptable for models.
As a temporal solution, I had to comment on those types which cause messages from lint-raml-cop.

- [ ] ramls/packages.raml should be fixed with https://issues.folio.org/browse/MODKBEKBJ-22
- [ ] ramls/resources.raml should be fixed with https://issues.folio.org/browse/MODKBEKBJ-29


ramls/traits/sortable.raml
```diff
@@ -0,0 +1,9 @@
+queryParameters:
+  sort:
+    displayName: Sort options
+    type: string
+    enum: [name, relevance]
```
andrii-paias: There is an issue with enum objects used as query parameters: 
enum objects that are generated from raml are generated like this:
```
public enum EholdingsPackagesGetSort {
  NAME("name"),

  RELEVANCE("relevance");

  private String name;

  EholdingsPackagesGetSort(String name) {
    this.name = name;
  }
}
```
RestVerticle then compares them with query parameters using case senstive comparison with value returned from toString(), so query parameters will be parsed correctly only if we send this in query "?sort=RELEVANCE", 
we can use String parameter instead of enum and then compare them manually properly, 
or maybe there is some other way around this issue

natali-zaitseva: agree with you, Andrii; violates the module contract that is why I changed the values inside raml files to string values and expanded 'example' property with possible values like in the following image
image

ramls/titles.raml
```diff
@@ -5,176 +5,75 @@ version: v1
...
         body:
           application/vnd.api+json:
-            example: !include examples/titles/titles_post_200_response.json
+            description: The server has successfully fulfilled the POST request.
+            type: title
+            example:
+              strict: false
+              value: !include examples/titles/titles_post_200_response.json
       400:
         description: Bad
```
sduvvuri-ebsco: Is this left incomplete/still WIP?

ramls/traits/filterable.raml
```diff
@@ -2,13 +2,15 @@ queryParameters:
...
     type: string
-    enum: ["all", "true", "false", "ebsco"]
     description: Filter to narrow down results based on selection status
+    example: |
+      ["all", "true", "false", "ebsco"]
     required: false
   filter[type]:
     displayName: Content type
     type: string
-    enum: ["all", "aggregatefulltext", "abstractandindex", "ebook", "ejournal", "print", "unknown", "onlinereference"]
```
sduvvuri-ebsco: why is this enum removed?

ramls/traits/includeResources.raml
```diff
@@ -0,0 +1,8 @@
+queryParameters:
+  include:
+    displayName: Related Resources
+    type: string
+    description: Include related resource obects, each representing a partnering of this title with a package.  Any bogus value, like `include=deciduousTrees`, will be silently ignored.
```
sduvvuri-ebsco: A really minor nitpick with the spelling of "objects"

ramls/types/proxy.json
```diff
@@ -16,5 +16,8 @@
       "description": "Whether this proxy is inherited",
       "example": true
     }
-  }
-}
\ No newline at end of file
+  },
+  "required": [
```
sduvvuri-ebsco: I created a separate proxy.json and proxyPut.json because we do not want user to be able to update "inherited". In mod-kb-ebsco, since UI passes it, we receive it and always set it to false before making a request to RM API. I thought in Java, we can just ignore the property "proxy.inherited" even if passed from UI. Is that possible?

ramls/types/resources/resourcesRelationships.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Metadata of relationships for a given object",
+  "description": "Metadata of relationships for a given object",
+  "javaType": "org.folio.rest.jaxrs.model.ResourceRelationships",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "vendor": {
```
sduvvuri-ebsco: we do not need "vendor" anymore. It is redundant with "provider". History is that earlier "vendor" in RM API was mapped to "vendor" in mod-kb but on usability testing request, we mapped "vendor" to "provider" and cleaned up "vendor" from mod-kb-ebsco but looks like its not cleaned completely. UI only recognizes "provider" and does not know "vendor". So, we can safely remove "vendor"

ramls/traits/filterable.raml
natali-zaitseva: because of the issue which was described by Andrii Piais above.
If we leave values as enum then:
> RestVerticle then compares them with query parameters using case senstive comparison with value returned from toString(), so query parameters will be parsed correctly only if we send this in query "?sort=RELEVANCE",

ramls/traits/includeResources.raml
natali-zaitseva: fixed

ramls/types/resources/resourcesRelationships.json
natali-zaitseva: the reason why "vendor" was added to the resourcesRelationships.json is that I have performed a request from the collection `../eholdings/providers/19?include=packages` and found following objects inside a relationship
image
@sduvvuri-ebsco correct me if I am wrong, we will have only two objects  "resources" and "provider", right?

ramls/titles.raml
natali-zaitseva: it was copied from original raml file 
image
but you are right, it can be left like this. I have noticed that for invalid POST user receives 422 error code instead of 400 as it is in raml file. 
image
I will change the raml to return 422 code and add an example section. 
Thank you @sduvvuri-ebsco for the notice.

sduvvuri-ebsco: 👍

ramls/traits/filterable.raml
sduvvuri-ebsco: Gotcha! Makes sense then to not have enum there and have them in examples like Andrii is doing in his PR

ramls/types/resources/resourcesRelationships.json
sduvvuri-ebsco: yes @natali-zaitseva , understand why you have it there in the first place and in mod-kb-ebsco-java, we can only return "resources" and "provider" and eliminate "vendor". Thanks!

PR #25, opened 2018-10-26
MODKBEKBJ-18 GET /eholdings/providers endpoint

andrii-paias:
## Purpose
Implement 
GET /eholdings/providers endpoint 
## Approach
Used mapping from ruby module as reference.
Used @Validated annotation to validate parameters, added Aspectj aspect to wrap validation errors into 
JsonApi error structure.
#### TODOS and Open Questions
- [x] Providers response has "relationships" attribute that has structure as shown below. It seems that this attribute is always empty for this endpoint. I'm not sure if we need to return something or if we should  just always return this empty structure.
```
"relationships": {
                "packages": {
                    "meta": {
                        "included": false
                    }
                }
            }
```
- [x] There is also "providerToken" attribute, java model has an object with 4 fields for this attribute, but RM API returns only one string, ruby module  just puts this string into providerToken.value:
```
"providerToken": {
                    "value": "tokenzzz"
                }
```


carolegodfrey:
@andrii-paias -- regarding `providerToken` attribute -- there is a difference between what RM API returns when returning a list of vendors vs. GETing the details for an individual vendor..

For example -- the below GET for an individual vendor shows the vendor token details with 4 fields populated. 

GET https://api.ebsco.io/rm/rmaccounts/apidvcorp/vendors/18

```
{
    "isCustomer": false,
    "packagesSelected": 9,
    "packagesTotal": 244,
    "vendorId": 18,
    "vendorName": "Gale | Cengage",
    "proxy": {
        "id": "EZProxy",
        "inherited": false
    },
    "vendorToken": {
        "factName": "[[galesiteid]]",
        "prompt": "/itweb/",
        "helpText": "<ul>\r\n    <li>Enter your Gale<sup>®</sup> site ID in the space provided below. The site ID may contain a combination of alpha/numeric characters, varying in length. <blockquote style=\"margin-right: 0px;\" dir=\"ltr\">\r\n    <p> Example: The site ID immediately follows /itweb/ in a URL. The site ID in the following URL is <i>aa11bb22</i>. </p>\r\n    </blockquote></li>\r\n</ul>\r\n<blockquote style=\"margin-right: 0px;\" dir=\"ltr\"><blockquote style=\"margin-right: 0px;\" dir=\"ltr\">\r\n<p><span style=\"text-decoration: underline;\">http://infotrac.galegroup.com/itweb/aa11bb22?db=AIM</span></p>\r\n</blockquote></blockquote><br />\r\n<ul>\r\n    <li>If no site ID is specified, your Gale Group links may not function properly, as Gale Group requires this information for authentication. </li>\r\n    <li>If you are unable to locate the site ID, please contact Gale Group. For contact information, visit: <a href=\"http://access.gale.com/authentication/\">http://access.gale.com/authentication/</a>. </li>\r\n</ul>\r\n",
        "value": "provider token test again"
    }
}
```

We will need this object for the case of 
https://issues.folio.org/browse/MODKBEKBJ-19

carolegodfrey:
Regarding relationships attribute - you are correct in that it is empty for this list endpoint. It is definitely used/populated when GETing provider details for an individual provider to get the list of associated packages for a provider (right hand panel in image below shows associated package list)

Detail parameter include=packages
GET eholdings/providers/18?include=packages

screen shot 2018-10-28 at 7 25 52 pm

src/test/java/org/folio/util/TestUtil.java
```diff
@@ -26,4 +32,21 @@ public static File getFile(String filename) throws URISyntaxException {
...
   }
+
+  /**
+   * Mocks wiremock server to return RM API configuration from specified file,
+   * RM API url will be changed to wiremockUrl so that following requests to RM API will be sent to wiremock instead
+   * @param configurationsFile configuration file, first config object must contain url config
+   * @param wiremockUrl wiremock url with port
+   */
+  public static void mockConfiguration(String configurationsFile, String wiremockUrl) throws IOException, URISyntaxException {
+    ObjectMapper mapper = new ObjectMapper();
```
carolegodfrey: Nice having this in a test util - sure this will be used for most tests!

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
```diff
@@ -0,0 +1,129 @@
...
+      .header(TOKEN_HEADER)
+      .when()
+      .get("eholdings/providers?q=e&page=1&sort=name")
+      .then()
+      .statusCode(200)
+      .body("meta.totalResults" , notNullValue())
+      .body("data[0].type", equalTo("providers"))
+      .body("data[0].id", notNullValue())
+      .body("data[0].attributes.name", notNullValue());
+  }
```
carolegodfrey: Just wondering if we should check actual sample results here instead of checking for noNullValue?

src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -97,6 +105,36 @@ public RMAPIService(String customerId, String apiKey, String baseURI, Vertx vert
...
+        throw new IllegalStateException("failed to encode query using UTF-8");
+      }
+      parameters.add("search=" + encodedQuery);
+    }
+    parameters.add("offset=" + page);
+    parameters.add("count=" + count);
+    parameters.add("orderby=" + determineSortValue(sort, q));
+
+    return this.getRequest(constructURL("vendors?" + String.join("&", parameters)), Vendors.class);
+  }
```
carolegodfrey: It may be helpful to have tests for RMAPIService layer itself?

andrii-paias: We can write tests for RMAPIService, but that will require a lot of mocking of vertx classes (HttpClient, HttpClientRequest etc.), if we need to test that correct request is constructed for RM API, then we can extract this logic and test it separately, like this - https://github.com/folio-org/mod-kb-ebsco-java/pull/25/commits/92b6e474e19ccdc79936782288df70428fbdb8a2

descriptors/ModuleDescriptor-template.json
```diff
@@ -119,14 +119,20 @@
       "displayName": "get RM API configuration status",
       "description": "Get RM API configuration status"
     },
+    {
+      "permissionName": "kb-ebsco.provider.get",
```
sduvvuri-ebsco: How do you feel about renaming this permissionName to "kb-ebsco.providerCollection.get" since it is a list of providers? We will have a separate endpoint to get a provider using provider id.

ramls/traits/queriable.raml
```diff
@@ -26,8 +26,6 @@ queryParameters:
   sort:
     displayName: Sort options
     type: string
-    enum: [name, relevance]
-    description: Option by which results are sorted
+    description: Option by which results are sorted. Possible values are relevance, name.
```
sduvvuri-ebsco: I remember there was a problem in using the enum generated code in Java although I do not know what exactly that is, but will specifying strings in enum alleviate that problem instead of removing it?

ramls/types/configuration/jsonapiError.json
```diff
@@ -0,0 +1,23 @@
+{
```
sduvvuri-ebsco: I think its beneficial to move this schema to ramls/types directory since I foresee this being referenced and used in several error responses for different endpoints, not just configuration.

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
```diff
@@ -0,0 +1,96 @@
...
+          rmapiConfiguration.getUrl(), vertxContext.owner());
+        return rmapiService.retrieveProviders(q, page, count, sort);
+      })
+      .thenAccept(vendors ->
+        asyncResultHandler.handle(Future.succeededFuture(GetEholdingsProvidersResponse
+        .respond200WithApplicationVndApiJson(converter.convert(vendors)))))
+      .exceptionally(e -> {
+        logger.error(GET_PROVIDERS_ERROR_MESSAGE, e);
+        asyncResultHandler.handle(Future.succeededFuture(GetEholdingsProvidersResponse
+          .status(500)
```
sduvvuri-ebsco: There are certain validations we do in the Ruby code and give 400 with an appropriate error message. Also, sometimes, when we get an error message from RM API, we pass that error message through so its useful to the end-user. Example: If the user makes a request with count=120, then we give a 400 with errors.title being `Parameter Count is outside the range 1-100`, where as here, it looks like we are giving a 500 for any exception.

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+      .get("eholdings/providers?q=e&page=1&sort=name")
+      .then()
+      .statusCode(200)
+      .body("meta.totalResults" , equalTo(expectedTotalResults))
+      .body("data[0].type", equalTo("providers"))
+      .body("data[0].id", equalTo(id))
+      .body("data[0].attributes.name", equalTo(name))
+      .body("data[0].attributes.packagesTotal", equalTo(packagesTotal))
+      .body("data[0].attributes.packagesSelected", equalTo(packagesSelected))
+      .body("data[0].attributes.supportsCustomPackages", equalTo(supportsCustomPackages));
```
sduvvuri-ebsco: Also, just curious, why are we not checking for "providerToken" in response even if it ends up being null?

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+    RestAssured.given()
+      .spec(spec).port(port)
+      .header(new Header(RestConstants.OKAPI_URL_HEADER, "http://localhost:8080"))
+      .header(TENANT_HEADER)
+      .header(TOKEN_HEADER)
+      .when()
+      .get("eholdings/providers?q=e&count=1000")
+      .then()
+      .statusCode(400)
+      .body("errors.first.title" , notNullValue());
```
sduvvuri-ebsco: Can we check for the specific error message instead of a notNullValue()?

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    RestAssured.given()
+      .spec(spec).port(port)
+      .header(new Header(RestConstants.OKAPI_URL_HEADER, wiremockUrl))
+      .header(TENANT_HEADER)
+      .header(TOKEN_HEADER)
+      .when()
+      .get("eholdings/providers?q=e&count=1")
+      .then()
+      .statusCode(500)
```
sduvvuri-ebsco: As mentioned in a comment above, not sure if we should treat all RM API error responses as 500's in our code since some of them might be informative for the end-user?

src/main/java/org/folio/rmapi/builder/VendorUrlBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.util.ArrayList;
+import java.util.List;
+
+public class VendorUrlBuilder {
+
+  private static final String VENDOR_NAME_PARAMETER = "VendorName";
+  private static final String RELEVANCE_PARAMETER = "Relevance";
```
sduvvuri-ebsco: I understand that RM API at the moment is case in-insensitive about these parameters being passed in request but when we play around with their docs, they always have queries in lower-case. Should we stick to that convention and make these all lower-case?

src/test/java/org/folio/rmapi/builder/VendorUrlBuilderTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Test
+  public void shouldBuildUrlForRelevanceSortWhenSortIsNotSetAndQueryIsSet(){
+    String path = new VendorUrlBuilder().q("higher education").build();
+    assertEquals("vendors?search=higher+education&offset=1&count=25&orderby=Relevance", path);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void shouldThrowExceptionIfSortValueIsInvalid() {
+    new VendorUrlBuilder().sort("abc").build();
```
sduvvuri-ebsco: Should we check for specific status code and error message in Rest Assured tests for this scenario?

sduvvuri-ebsco:
Looks pretty good overall @andrii-paias , I am a little concerned if we are passing the correct error responses and status codes from mod-kb-ebsco-java. Except that, this looks pretty good 👍

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
andrii-paias: There is validation of parameters in mod-kb-ebsco-java module, so if parameter is invalid mod-kb-ebsco-java will imediately return a 400 error, without sending request to RM API. 
Validation is implemented by using @Validate annotation, parameters will be automatically validated against interface generated from raml (e.g. count parameter is generated like this "@QueryParam("count") @DefaultValue("25") @Min(0) @Max(100) int count", so if count is greater than 100 an error will be returned),
here's an example error
```
{
    "errors": [
        {
            "detail": "\n 'count' parameter is incorrect. parameter value {1001} is not valid: must be less than or equal to 100"
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}
```

But it is still nice to return RM API error message in case of other errors.
I changed RMAPIService to include response body from RM API. That way it will be included in 500 response:
```
{
    "errors": [
        {
            "detail": "Unauthorized Access to https://sandbox.ebsco.io/rm/rmaccounts/apidvgvmt/vendors?search=Cengage&offset=1&count=10&orderby=relevance, response body =\"{\"message\":\"Unauthorized\"}\n\""
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}
```

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
andrii-paias: Error message is constructed by code in RMB module, I think if we test for exact error String then the test will be brittle, in the future exact minimum/maximum constraints can change, or RMB might  phrase the error message differently

src/test/java/org/folio/rmapi/builder/VendorUrlBuilderTest.java
andrii-paias: Invalid sort parameter will result in 500 error,
I moved validation of sort parameter to EholdingsProvidersImpl so that 400 error is returned.

andrii-paias:
Thanks for reviewing this PR, Sobha,
updated PR with some changes to how errors are handled

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
sduvvuri-ebsco: Is there a way we can customize the error messages we want to return using @validate annotation? The reason I ask is, in mod-kb-ebsco, we return certain customized error messages and populate them in errors.title which is what is displayed in UI as a toast message.
Also, regarding 500's, I think it will be good to differentiate status codes based on type of exceptions and return them appropriately, like 403 probably for a "unauthorized access" and again, not sure if that is the exact error message being returned from mod-kb. Thanks!

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
sduvvuri-ebsco: Per my comment above, is there a way we can pass our customized error message to RMB for it to pass through?

sduvvuri-ebsco:
Thank you for addressing most of the comments on this PR @andrii-paias. Looks pretty good and again, this is not specific to this PR but the outstanding comments are in general things to be figured out with error message handling in our code and RMB. I am OK if we want to deal with this later in another user story/another PR but as I mentioned UI displays those specific error messages to end-users and we want to make sure that they are concise and convey the point. If you want us to deal with this in another PR, please create a user story to address error message handling in mod-kb-ebsco-java and we can merge this PR. Thanks!

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
dmtkachenko: let's address this in another US

PR #27, opened 2018-11-02
MODKBEKBJ-35 Implement GET /eholdings/titles endpoint

andrii-paias:
## Purpose
Implement
GET /eholdings/titles endpoint

## Approach
Used mapping from ruby module as reference.

#### TODOS and Open Questions
- [x] Ruby module has subjects.subject, and subject.name attributes, they seem to contain same string:
```
   "subjects": [
                    {
                        "type": "BISAC",
                        "name": "BIOGRAPHY & AUTOBIOGRAPHY / General",
                        "subject": "BIOGRAPHY & AUTOBIOGRAPHY / General"
                    },
```
raml schema only has subjects.subject attribute,
do we need to add name attribute as well?
ruby module doesn't explicitly return name, it just returns all of the attributes returned from RM API

## Screenshots
java_1


sduvvuri-ebsco:
Good question @andrii-paias , your observation is correct in that Ruby module passes through response from RM API which contains "name" and "subject" having same values. We notified them of it and they have a story in their backlog to get rid of "name" which is why we did not add it in the JSON schema and added only "subject". Hope this helps.

ramls/traits/sortable.raml
```diff
@@ -4,5 +4,4 @@ queryParameters:
     type: string
     description: Option by which results are sorted. Defaults to relevance if query or name if no query. Possible values are name, relevance.
     example : name
-    default: relevance
```
sduvvuri-ebsco: curious why this is removed from raml?

src/main/java/org/folio/rest/aspect/HandleValidationErrors.java
```diff
@@ -0,0 +1,19 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+/**
+ * If method annotated by @HandleValidationErrors throws ValidationException
+ * exception will be caught and 400 error will be returned with exception message
+ * wrapped into json according to jsonapi.
+ */
+@Target( { ElementType.METHOD })
```
sduvvuri-ebsco: A really minor nitpick on the spacing in the above annotation

src/main/java/org/folio/rest/converter/TitleConverter.java
```diff
@@ -0,0 +1,104 @@
...
+
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class TitleConverter {
+
+  private static final TitleRelationShip EMPTY_PACKAGE_RELATIONSHIP = new TitleRelationShip()
```
sduvvuri-ebsco: Do we want to rename this to EMPTY_RESOURCES_RELATIONSIP instead?

ramls/traits/sortable.raml
andrii-paias: there was a very minor problem:
ruby module returns correct response for this path
/eholdings/providers
but returns an error for this path
/eholdings/providers?sort=relevance

if default is set to relevance then we can't distinguish between those two requests.

now that I think about it, maybe it's better to just ignore sort parameter in case of
/eholdings/providers?sort=relevance
and treat it as /eholdings/providers

what do you think?

src/main/java/org/folio/rest/converter/TitleConverter.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private static final Map<Integer, String> IDENTIFIER_SUBTYPES = new HashMap<>();
+  static {
+    IDENTIFIER_SUBTYPES.put(1,"Print");
+    IDENTIFIER_SUBTYPES.put(2,"Online");
+  }
+
+  public TitleCollection convert(Titles titles) {
+    List<org.folio.rest.jaxrs.model.Titles> providerList = titles.getTitleList().stream()
```
sduvvuri-ebsco: confused why this is named "providerList"?

src/main/java/org/folio/rest/impl/EholdingsTitlesImpl.java
```diff
@@ -0,0 +1,110 @@
...
+
+import javax.ws.rs.core.Response;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+public class EholdingsTitlesImpl implements EholdingsTitles {
+
+  private static final String GET_TITLES_ERROR_MESSAGE = "Failed to retrieve titles";
+
+  private final Logger logger = LoggerFactory.getLogger(EholdingsConfigurationImpl.class);
```
sduvvuri-ebsco: this logger should be getLogger(EholdingsTitlesImpl.class) instead?

src/main/java/org/folio/rest/impl/EholdingsTitlesImpl.java
```diff
@@ -0,0 +1,110 @@
...
+    this(
+      new RMAPIConfigurationServiceCache(
+        new RMAPIConfigurationServiceImpl(new ConfigurationClientProvider())),
+      new HeaderValidator(),
+      new TitleParametersValidator(),
+      new TitleConverter());
+  }
+
+  public EholdingsTitlesImpl(RMAPIConfigurationService configurationService,
+                                HeaderValidator headerValidator, TitleParametersValidator parametersValidator,
```
sduvvuri-ebsco: spacing seems a little off here?

src/main/java/org/folio/rest/impl/EholdingsTitlesImpl.java
```diff
@@ -0,0 +1,110 @@
...
+            .header("Content-Type", "application/vnd.api+json")
+            .entity(ErrorUtil.createErrorFromRMAPIResponse(rmApiException))
+            .build()));
+        }
+        else {
+          asyncResultHandler.handle(Future.succeededFuture(Response
+            .status(500)
+            .header("Content-Type", "application/vnd.api+json")
+            .entity(ErrorUtil.createError(e.getCause().getMessage()))
+            .build()));
```
sduvvuri-ebsco: we could probably extract the code in "if" and "else" into a method and just pass the status code as param to that method since code is duplicated?

src/main/java/org/folio/rest/validator/TitleParametersValidator.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class TitleParametersValidator {
+
+  private static final List<String> SELECTED_VALUES = Arrays.asList("true", "false", "ebsco");
+  private static final List<String> TYPE_VALUES = Arrays.asList("audiobook", "book", "bookseries", "database", "journal", "newsletter", "newspaper", "proceedings", "report", "streamingaudio", "streamingvideo", "thesisdissertation", "website", "unspecified");
+
+  /**
+   * @throws ValidationException if validation fails
+   */
+  public void validate(String filterSelected, String filterType, String filterName, String filterIsxn, String filterSubject,
```
sduvvuri-ebsco: I am curious about something here: 
1. From UI, we are passing filter[selected] - correct? how is that getting converted to filterSelected?

src/main/java/org/folio/rest/validator/TitleParametersValidator.java
```diff
@@ -0,0 +1,42 @@
...
+      .filter(Objects::nonNull)
+      .count();
+    if(nonNullFilters > 1){
+      throw new ValidationException("Conflicting filter parameters");
+    }
+    if(searchParameters.stream()
+      .anyMatch(""::equals)){
+      throw new ValidationException("Required parameter 'search' is missing.");
+    }
+    if((sort != null && !Sort.contains(sort.toUpperCase()))){
```
sduvvuri-ebsco: This might not be the appropriate place for this comment but one observation is that in mod-kb, by default the sort is expected to be "relevance" and UI does not pass it if "relevance" is selected as sort parameter. Backend code defaults it to "relevance" if UI does not provide anything for sort

src/main/java/org/folio/rmapi/builder/QueriableUrlBuilder.java
```diff
@@ -52,19 +56,20 @@ public String build(){
...
-      return query == null ? VENDOR_NAME_PARAMETER : RELEVANCE_PARAMETER;
+      return query == null ? nameParameter : RELEVANCE_PARAMETER;
     }
     switch (sort){
       case RELEVANCE:
         return RELEVANCE_PARAMETER;
       case NAME:
-        return VENDOR_NAME_PARAMETER;
+        return nameParameter;
+      default:
```
sduvvuri-ebsco: Here, I think we should change the default to be "relevance" instead of throwing an invalid sort param

src/main/java/org/folio/rmapi/model/CustomerResources.java
```diff
@@ -0,0 +1,160 @@
+package org.folio.rmapi.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class CustomerResources {
```
sduvvuri-ebsco: I am a little curious about this model - is this related to this documentation here? https://s3.amazonaws.com/foliodocs/api/mod-kb-ebsco/eholdings.html#eholdings_resources__resourceid__get

src/main/java/org/folio/rmapi/model/CustomerResources.java
```diff
@@ -0,0 +1,160 @@
...
+
+  @JsonProperty("titleId")
+  private String titleId;
```
sduvvuri-ebsco: If this model is derived from the documentation mentioned above, I see that in sample response - titleId is an integer and not a string?

src/main/java/org/folio/rmapi/model/CustomerResources.java
```diff
@@ -0,0 +1,160 @@
...
+  @JsonProperty("packageName")
+  private String packageName;
+
+  @JsonProperty("isPackageCustom")
+  private Boolean isPackageCustom;
+
+  @JsonProperty("vendorId")
+  private Integer vendorId;
+
+  @JsonProperty("vendorName")
```
sduvvuri-ebsco: Both these "vendorId" and "vendorName" can be removed. These will always be redundant with "providerId" and "providerName" and UI will only recognize the latter. Ruby code should be cleaned up to not give "vendor*" in its responses but we will deal with that cleaning part here.

src/main/java/org/folio/rmapi/model/CustomerResources.java
```diff
@@ -0,0 +1,160 @@
...
+  @JsonProperty("managedEmbargoPeriod")
+  private EmbargoPeriod managedEmbargoPeriod;
+
+  @JsonProperty("customEmbargoPeriod")
+  private EmbargoPeriod customEmbargoPeriod;
+
+  @JsonProperty("url")
+  private String url;
+
+  @JsonProperty("userDefinedField1")
```
sduvvuri-ebsco: Similarly these "userDefinedFields" are not in the above sample response in documentation - do we get them in requests to mod-kb(Ruby)? If yes, then probably that doc is incomplete?

src/main/java/org/folio/rmapi/model/Error.java
```diff
@@ -0,0 +1,26 @@
...
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Error {
+  @JsonProperty("code")
+  private String code;
+  @JsonProperty("message")
+  private String message;
+  @JsonProperty("subCode")
+  private String subCode;
+
+  public String getCode() {
```
sduvvuri-ebsco: I don't think we are interested in "code" and "subCode" returned by RM API - just the error message in some cases

src/test/java/org/folio/rest/impl/EholdingsTitlesTest.java
```diff
@@ -0,0 +1,150 @@
...
+      .build();
+  }
+
+  @Before
+  public void setUp() {
+    RMAPIConfigurationCache.getInstance().invalidate();
+  }
+
+  @Test
+  public void shouldReturnProvidersOnGet() throws IOException, URISyntaxException {
```
sduvvuri-ebsco: shouldReturnTitlesOnGet() ?

src/main/java/org/folio/rmapi/model/CustomerResources.java
andrii-paias: This model corresponds to response from RM API, 
I've taken it from here 
https://developer.ebsco.com/interactive/holdingsiq#/

src/main/java/org/folio/rmapi/builder/QueriableUrlBuilder.java
andrii-paias: Here sort is guaranteed to be not null, because there is a null check above,
"default:" part is included just to make java code compile, basically "default:" will only be executed if in the future someone adds more possible Sort enum values

src/main/java/org/folio/rest/validator/TitleParametersValidator.java
andrii-paias: yes, we're passing filter[selected] I think it is mapped to filterSelected correctly because filterSelected  is annotated with  @QueryParam("filter[selected]")  in the generated interface

src/main/java/org/folio/rmapi/model/Error.java
andrii-paias: RM API returns code and subcode, here's example
```
{
  "errors": [
    {
      "code": "1006",
      "message": "Parameter 'count' is outside the range 1-100.",
      "subCode": "0"
    }
  ]
}
```
it is taken from here 
https://developer.ebsco.com/interactive/holdingsiq#/

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
```diff
@@ -73,26 +74,41 @@ public void getEholdingsProviders(String q, String sort, int page, int count, Ma
...
         logger.error(GET_PROVIDERS_ERROR_MESSAGE, e);
-        asyncResultHandler.handle(Future.succeededFuture(GetEholdingsProvidersResponse
-          .status(500)
-          .header("Content-Type", "application/vnd.api+json")
-          .entity(ErrorUtil.createError(e.getCause().getMessage()))
-          .build()));
+        if(e.getCause() instanceof RMAPIServiceException){
+          RMAPIServiceException rmApiException = (RMAPIServiceException)e.getCause();
+          asyncResultHandler.handle(Future.succeededFuture(Response
+            .status(rmApiException.getRMAPICode())
```
sduvvuri-ebsco: I do not remember exactly what status codes RM API gives - are those regular HTTP status codes or internal codes of RM API?

sduvvuri-ebsco: Based on the above response, not sure if we should use RM API Code here to return status

andrii-paias: yes, it's http status code

src/main/java/org/folio/rmapi/builder/QueriableUrlBuilder.java
andrii-paias: yes

ramls/traits/sortable.raml
carolegodfrey: Took a look at specific error we are seeing in the case of `/eholdings/providers?sort=relevance`
It is returning the error from rm api
`Parameter OrderBy should be by ProviderName when parameter Search is empty.` 
and from RM API documentation -- `Search Term - When searching for vendors without a search parameter, sort options will be by vendorname. When the search parameter is not null, the default sort is by relevance.`
So default sort will depend on the presence of a q parameter

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
```diff
@@ -53,12 +55,11 @@ public EholdingsProvidersImpl(RMAPIConfigurationService configurationService,
...
-    if (!headerValidator.validate(okapiHeaders, asyncResultHandler)) {
-      return;
-    }
+    headerValidator.validate(okapiHeaders);
     if(sort != null && !Sort.contains(sort.toUpperCase())){
-      throw new ValidationException("sort parameter must have value \"relevance\" or \"name\"");
+      throw new ValidationException("Invalid sort parameter");
     }
     Sort nameSort = sort != null ? Sort.valueOf(sort.toUpperCase()) : null;
     CompletableFuture.completedFuture(null)
```
carolegodfrey: If sort is null -- we should apply logic to set it base on whether or not query string q has a value. Below links in ruby code show this logic for providers and titles both. If query string is null -- sort should default to relevance

https://github.com/folio-org/mod-kb-ebsco/blob/f4a8c7e6a63373a24cb78ff277c7b8e65492a2c3/app/models/title.rb#L61

https://github.com/folio-org/mod-kb-ebsco/blob/f4a8c7e6a63373a24cb78ff277c7b8e65492a2c3/app/models/provider.rb#L21

src/main/java/org/folio/rmapi/model/CustomerResources.java
carolegodfrey: Believe vendorId and vendorName are ok in this context since this as how the RM API response refers to these fields which in turn get mapped to provider

src/main/java/org/folio/rmapi/model/Error.java
carolegodfrey: I think its ok if the codes appear in these objects since they are used to serialize the results from rm api response. When the error is built from rm api response (in ErrorUtil.java) we are only using the message

src/test/java/org/folio/rest/impl/EholdingsTitlesTest.java
```diff
@@ -0,0 +1,150 @@
...
+      .body("data[0].attributes.subjects[0].subject", equalTo("BUSINESS & ECONOMICS / Small Business"))
+
+      .body("data[0].attributes.identifiers[0].id", equalTo("7209484"))
+      .body("data[0].attributes.identifiers[1].id", equalTo("978-1-74216-894-4"))
+      .body("data[0].attributes.identifiers[2].id", equalTo("978-0-7303-7792-4"))
+
+      .body("data[0].attributes.identifiers[0].type", equalTo("ISSN"))
+      .body("data[0].attributes.identifiers[1].type", equalTo("ISBN"))
+      .body("data[0].attributes.identifiers[2].type", equalTo("ISBN"))
+
```
carolegodfrey: Might be helpful to have a comment or even a separate test -- noting that identifiers get sorted by type and subtype. This is happening - here but not apparent from the test

src/test/java/org/folio/rest/impl/EholdingsTitlesTest.java
```diff
@@ -0,0 +1,150 @@
...
+      .spec(spec).port(port)
+      .header(new Header(RestConstants.OKAPI_URL_HEADER, wiremockUrl))
+      .header(TENANT_HEADER)
+      .header(TOKEN_HEADER)
+      .when()
+      .get("eholdings/titles?filter[name]=news")
+      .then()
+      .statusCode(500);
+  }
+}
```
carolegodfrey: Might also be helpful to have a mock of an rm api error response and show how it is being mapped in terms of mod-kb-ebsco error respponse

src/main/java/org/folio/rest/impl/EholdingsTitlesImpl.java
dmtkachenko: @sduvvuri-ebsco this is still not corrected

src/main/java/org/folio/rest/impl/EholdingsTitlesImpl.java
sduvvuri-ebsco: Now that I look at the code, I realize that the type of error response that we are using from ErrorUtil differs, one uses `ErrorUtil.createErrorFromRMAPIResponse` and the other uses `ErrorUtil.createError` with different exceptions being passed in. So, I think its fine the way it is.

src/main/java/org/folio/rmapi/model/CustomerResources.java
sduvvuri-ebsco: You are right Carole, I just realized that this class is mapping RM API json props.

src/main/java/org/folio/rmapi/model/CustomerResources.java
sduvvuri-ebsco: This comment can be ignored too given that this is coming from RM API.

src/main/java/org/folio/rmapi/model/Error.java
sduvvuri-ebsco: sounds good 👍

sduvvuri-ebsco:
I do not mind changing it to use Objects.nonNull instead of object != null but under the hood, Objects.nonNull is eventually checking for whether that object is null or not using !=, its a style preference, do you have a strong opinion about one style over other @natali-zaitseva ?

sduvvuri-ebsco:
Also, unsure about which line specifically you are suggesting using StringUtils::isEmpty? Could you please clarify? Thanks!

sduvvuri-ebsco:
@natali-zaitseva : I looked at using StringUtils.isEmpty instead of ""::equals and observed the difference. StringUtils.isEmpty returns true even if null value is passed where ""::equals explicitly checks for empty string. There is a difference in that Andrii is checking that the value passed for a given filter cannot be empty but that for example: if filter[name] is passed in, then the other filters filter[isxn] etc. can be null

natali-zaitseva:
Thanks for checking @sduvvuri-ebsco. Then, we can skip my comment about StringUtils in this case. The pull request looks good to me

PR #28, opened 2018-11-02
MODKBEKBJ-19 - Implement GET /eholdings/providers/{provider_id}

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-19 we want to have the ability to get provider by id by supporting '/eholdings/providers/{provider_id}' endpoint

## Approach
_How does this change fulfill the purpose?_
 * Implement get providers by id method.
* Map vendor returned from RM API to a single Provider
* Add permissions to the request

#### TODOS and Open Questions
- [ ] update the code to support parameter '?include=packages' after packages story is done.

### DIFF
--- a/src/main/java/org/folio/rest/converter/VendorConverter.java
+++ b/src/main/java/org/folio/rest/converter/VendorConverter.java
@@ -1,15 +1,10 @@
 package org.folio.rest.converter;
 
-import org.folio.rest.jaxrs.model.MetaDataIncluded;
-import org.folio.rest.jaxrs.model.MetaTotalResults;
-import org.folio.rest.jaxrs.model.Packages;
-import org.folio.rest.jaxrs.model.ProviderCollection;
-import org.folio.rest.jaxrs.model.ProviderListDataAttributes;
-import org.folio.rest.jaxrs.model.Providers;
-import org.folio.rest.jaxrs.model.Relationships;
-import org.folio.rest.jaxrs.model.Token;
+import org.folio.rest.jaxrs.model.*;
 import org.folio.rest.util.RestConstants;
 import org.folio.rmapi.model.Vendor;
+import org.folio.rmapi.model.VendorById;
+import org.folio.rmapi.model.VendorToken;
 import org.folio.rmapi.model.Vendors;
 
 import java.util.List;
@@ -18,9 +13,11 @@ import java.util.stream.Collectors;
 public class VendorConverter {
 
   private static final Relationships EMPTY_PACKAGE_RELATIONSHIP = new Relationships()
-    .withPackages(new Packages().withMeta(
-      new MetaDataIncluded()
-            .withIncluded(false)));
+    .withPackages(new Packages()
+      .withMeta(new MetaDataIncluded()
+        .withIncluded(false))
+      .withData(null));
+  private final String PROVIDERS_TYPE = "providers";
 
   public ProviderCollection convert(Vendors vendors) {
     List<Providers> providerList = vendors.getVendorList().stream()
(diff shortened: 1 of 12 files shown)


sduvvuri-ebsco:
On your open question, please create a separate user story for "update the code to support parameter '?include=packages' after packages story is done" in backlog. Thanks!

src/main/java/org/folio/rest/converter/VendorConverter.java
```diff
@@ -45,4 +41,25 @@ private Providers convertVendor(Vendor vendor) {
...
+    return new Provider()
+      .withData(new ProviderData()
+        .withId(String.valueOf(vendor.getVendorId()))
+        .withType(PROVIDERS_TYPE)
+        .withAttributes(new ProviderDataAttributes()
+          .withName(vendor.getVendorName())
+          .withPackagesTotal(vendor.getPackagesTotal())
+          .withPackagesSelected(vendor.getPackagesSelected())
+          .withSupportsCustomPackages(vendor.isCustomer())
+          .withProviderToken(token != null ? new Token().withValue(token) : null)
```
sduvvuri-ebsco: Here for "providerToken" object - it has the following attributes - "factName", "prompt", "helpText" and "value" and we want to populate all - not just value - this is a bit different than "providers" endpoint. Unfortunately, RM API does not give consistent results for collection and detailed record in all cases.

sduvvuri-ebsco: To observe the difference mentioned above - you can play around with GET /custid/vendors and GET /custid/vendors/vendorId in this interactive docs - https://developer.ebsco.com/interactive/holdingsiq

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
```diff
@@ -84,7 +89,33 @@ public void getEholdingsProviders(String q, String sort, int page, int count, Ma
...
+      })
+      .thenAccept(vendor ->
+        asyncResultHandler.handle(Future.succeededFuture(GetEholdingsProvidersByProviderIdResponse
+          .respond200WithApplicationVndApiJson(converter.convertToVendor(vendor)))))
+      .exceptionally(e -> {
+        if (e.getCause() instanceof RMAPIResourceNotFoundException) {
+          asyncResultHandler.handle(Future.succeededFuture(GetEholdingsProvidersByProviderIdResponse
+            .respond404WithApplicationVndApiJson(ErrorUtil.createError(GET_PROVIDER_NOT_FOUND_MESSAGE))));
+        } else {
+          logger.error(GET_PROVIDERS_ERROR_MESSAGE, e);
```
sduvvuri-ebsco: I am unsure what GET_PROVIDERS_ERROR_MESSAGE is but does it have anything specific to a collection as opposed to an individual provider because then we might want to use a separate error message/tweak existing one to be generic

src/main/java/org/folio/rest/converter/VendorConverter.java
natali-zaitseva: thank you for review Sobha. Fixed it by changing  "vendorToken" property from String to Object including "factName", "prompt", "helpText" and "value" parameters.
screen shot 2018-11-05 at 2 34 25 pm
screen shot 2018-11-05 at 2 35 32 pm

sduvvuri-ebsco: 👍

natali-zaitseva:
> On your open question, please create a separate user story for "update the code to support parameter '?include=packages' after packages story is done" in backlog. Thanks!

@sduvvuri-ebsco story is created - https://issues.folio.org/browse/MODKBEKBJ-53

PR #29, opened 2018-11-05
MODKBEKBJ-22 Adjust RAML for Package endpoint to conform to JSON API

dmtkachenko:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-22, we want to tweak RAML and put together JSON schema that supports requests/responses for various /eholdings/packages endpoints conforming to JSON API schema.

## Approach
- organize raml directory
- tweak json schema for the following endpoints:
    - GET and POST for /eholdings/packages
    - GET, PUT, DELETE for /eholdings/packages/{packageId}
    - GET for /eholdings/packages/{packageId}/resources


ramls/traits/queriable.raml
```diff
@@ -4,4 +4,4 @@ queryParameters:
     type: string
     description: String used to search to retrieve a collection
     example: ABC-CLIO
-    required: true
+    required: false
```
sduvvuri-ebsco: @dmtkachenko : Just curious, why is this made false? Isn't a search param needed to be able to search a collection ?

ramls/packages.raml
```diff
@@ -1,169 +1,111 @@
...
+              strict: false
+              value: !include examples/packages/packages_post_422_response.json
   /{packageId}:
+    displayName: Get package by package Id
+    description: Instance of a package given packageId
+    uriParameters:
+        packageId:
+          description: identifier of the package formed from Provider Id and Package Id (see example)
+          type: string
+          example: 583-4345
```
sduvvuri-ebsco: Do we want to make packageId required: true?

ramls/types/packages/packageRelationships.json
```diff
@@ -0,0 +1,25 @@
...
+  "javaType": "org.folio.rest.jaxrs.model.PackageRelationship",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "resources": {
+      "type": "object",
+      "description": "Displays Resources Relationship if any",
+      "$ref": "../metaIncluded.json"
+    },
+    "vendor": {
```
sduvvuri-ebsco: We can eliminate "vendor" from here. "vendor" was replaced by "provider" and probably, in some parts of Ruby code, we did not clean up "vendor" completely. UI only uses "provider" and has no knowledge of "vendor"

ramls/types/packages/packagePutDataAttributes.json
```diff
@@ -0,0 +1,62 @@
...
+    },
+    "visibilityData": {
+      "type": "object",
+      "description": "Visibility data",
+      "$ref": "../visibilityData.json"
+    },
+    "packageToken": {
+      "type": "object",
+      "description": "Tokens are variables in content URLs that identify the customer. The token is text within the URL that needs to be replaced with an institute-specific value.",
+      "$ref": "../token.json"
```
sduvvuri-ebsco: We might want to replace the reference to use "tokenPut.json" instead of "token.json" due to the limited subset of properties that the user is allowed to update.

ramls/types/packages/packagePutDataAttributes.json
```diff
@@ -0,0 +1,62 @@
...
+    },
+    "proxy": {
+      "type": "object",
+      "description": "Proxy",
+      "$ref": "../proxy.json"
```
sduvvuri-ebsco: we might want to replace the reference to use "proxyPut.json" instead of "proxy.json" due to the limited subset of properties that the user is allowed to update.

ramls/traits/queriable.raml
dmtkachenko: it can be omitted. for instance you can apply filter but do not use "q" param:
screenshot_1

ramls/packages.raml
dmtkachenko: it's a uri parameter. w/o it it doesn't make sense at all
we do not define uri params as required explicitly in other ramls. see providerId from providers.raml for example:
```
/{providerId}:
    displayName: Get provider by provider Id
    description: Instance of a provider given providerId
    uriParameters:
      providerId:
        description: Provider Id of provider to get
        type: string
    get:
      description: Get provider given providerId
      is: [includable]
      responses:
        200:
..
```

ramls/traits/queriable.raml
sduvvuri-ebsco: 👍

PR #30, opened 2018-11-07
MODKBEKBJ-50 - Verify configuration setup

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-50 we need to change the behavior of backend validation if the configuration is not currently set up. 
For the situation when we do not have a configuration entity and perform 
`GET /status`
we recieve following response 
`Internal server error` with code 500
## Approach
Added additional check for RMAPI configuration parameters.
This check will prevent retriving configuration details if configuration credentials are empty.
In case we do not have configuration and and perform 
`GET /status`
we will recieve 
`{
   “data”: {
       “id”: “status”,
       “type”: “statuses”,
       “attributes”: {
           “isConfigurationValid”: false
       }
   },
   “jsonapi”: {
       “version”: “1.0”
   }
}`
with response code 200 
#### TODOS and Open Questions
- [ ] There is should be UI story for behavior investigation when backend returns  `  “isConfigurationValid”: false` because for the current moment seems UI does not provide option to set up configuration
screen shot 2018-11-06 at 5 34 46 pm

## Learning
there is the issue with maven-surface-plugin, which brokes Jenkins build and already reported https://issues.folio.org/browse/FOLIO-1609 and https://issues.apache.org/jira/browse/SUREFIRE-1588

### DIFF
--- a/src/main/java/org/folio/config/RMAPIConfigurationServiceCache.java
+++ b/src/main/java/org/folio/config/RMAPIConfigurationServiceCache.java
@@ -44,8 +44,6 @@ public class RMAPIConfigurationServiceCache implements RMAPIConfigurationService
   public CompletableFuture<Boolean> verifyCredentials(RMAPIConfiguration rmapiConfiguration, Context vertxContext) {
     if(rmapiConfiguration.getConfigValid() != null){
       return CompletableFuture.completedFuture(rmapiConfiguration.getConfigValid());
-    } else if(isConfigurationParametersInvalid(rmapiConfiguration)){
-      return CompletableFuture.completedFuture(false);
     }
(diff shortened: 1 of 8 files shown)


natali-zaitseva:
@carolegodfrey @sduvvuri-ebsco I have changed the response of 'eholdings/configuration' when there is no configuration to 

`{
    "data": {
        "id": "configuration",
        "type": "configurations",
        "attributes": {
            "rmapiBaseUrl": "https://sandbox.ebsco.io"
        }
    },
    "jsonapi": {
        "version": "1.0"
    }
}`

screen shot 2018-11-07 at 6 28 32 pm

values of 'customerId' and 'apiKey' are null,  that is why they are not included into a response. If the absence of those values violates the contract for UI, please notify me.

PR #37, opened 2018-11-13
MODKBEKBJ-27: delete package

andrii-paias:
## Purpose
Implement delete package endpoint

Moved common logic for tests to separate abstract base class

Added validation for case when provider query is empty - this is not related to this story, but this change seems too small for a separate PR

## Approach
Send PUT request to RM API with body containing "isSelected": false

Changes in logic are in this commit - https://github.com/folio-org/mod-kb-ebsco-java/pull/37/commits/64c7b9efde1b25741b053fc4124423715d4183fe
Other changes are mostly for test refactoring


src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
```diff
@@ -74,6 +74,9 @@ public void getEholdingsProviders(String q, String sort, int page, int count, Ma
     if(!Sort.contains(sort.toUpperCase())){
       throw new ValidationException("Invalid sort parameter");
     }
+    if("".equals(q)){
+      throw new ValidationException("Search parameter cannot be empty");
+    }
     CompletableFuture.completedFuture(null)
```
carolegodfrey: I'm not so sure we want an error in this case  - I can make a requests such as `GET eholdings/providers` against the Ruby module and I get a list of providers sorted by provider name (vendorname is rm api)

src/test/java/org/folio/rest/impl/WireMockTestBase.java
```diff
@@ -0,0 +1,69 @@
...
+import org.folio.util.TestUtil;
+import org.junit.Before;
+import org.junit.BeforeClass;
+
+/**
+ * Base test class for tests that use wiremock and vertx http servers,
+ * test that inherits this class must use VertxUnitRunner as
+ */
+public abstract class WireMockTestBase {
+  protected static final Header CONTENT_TYPE_HEADER = new Header("Content-Type", "application/vnd.api+json");
```
carolegodfrey: Nice to have this!

src/test/java/org/folio/rest/impl/EholdingsPackagesImplTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+    RequestSpecification requestSpecification = getRequestSpecification();
+
+    RestAssured.given()
+      .spec(requestSpecification)
+      .when()
+      .delete("eholdings/packages/" + providerId + "-" + packageId)
+      .then()
+      .statusCode(HttpStatus.SC_NO_CONTENT);
+  }
```
carolegodfrey: Could we also add a  WireMock.verify for the PUT request to RM API? Or is that considered redundant with the stub for the put above?

andrii-paias: Thanks, added explicit verification for that

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
andrii-paias: in case of "eholdings/providers" q will be null and it will work ok, 
this is the case of "eholdings/providers?q=", for this Ruby module returns an error, and there are Postman tests that check for this error

carolegodfrey: ahh -- got it -- thanks for clarification

sduvvuri-ebsco: Isn't this the error message that we changed to "filter[*] cannot be empty" - If we still have "search parameter cannot be empty" - we might want to change that to filter[*] cannot be empty because "search" is a RM API term.

src/main/java/org/folio/rmapi/model/PackageData.java
```diff
@@ -0,0 +1,108 @@
...
+  @JsonProperty("isTokenNeeded")
+  private Boolean isTokenNeeded;
+  @JsonProperty("allowEbscoToAddTitles")
+  private Boolean allowEbscoToAddTitles;
+  @JsonProperty("packageType")
+  private String packageType;
+  @JsonProperty("proxy")
+  private Proxy proxy;
+  @JsonProperty("packageToken")
+  private VendorToken packageToken;
```
sduvvuri-ebsco: Are we using class VendorToken here because structure of both vendorToken and packageToken is the same or is it a typo?

src/test/java/org/folio/rest/impl/EholdingsPackagesImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+    RequestSpecification requestSpecification = getRequestSpecification();
+
+    RestAssured.given()
+      .spec(requestSpecification)
+      .when()
+      .delete("eholdings/packages/" + providerId + "-" + packageId)
+      .then()
+      .statusCode(HttpStatus.SC_BAD_REQUEST);
+  }
+}
```
sduvvuri-ebsco: can we also add one more test to see that a GET of a deleted custom package results in a 404 Not found with error.title = "Package not found"

src/main/java/org/folio/rmapi/model/PackageData.java
andrii-paias: Yes, in RM API docs it is the same object with same structure, still it seems strange to have this class named VendorToken, renamed it to TokenInfo

src/test/java/org/folio/rest/impl/EholdingsPackagesImplTest.java
andrii-paias: get package by id endpoint is not implemented yet, we will need to add this after we implement it

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
andrii-paias: The parameter name is "q", we can change error message to say "q", I just thought that if at some point this message will be shown on UI "search" will be easier to understand

src/main/java/org/folio/rmapi/exception/RMAPIResourceNotFoundException.java
```diff
@@ -4,11 +4,12 @@
  * @author cgodfrey
  *
  */
-public class RMAPIResourceNotFoundException extends RMAPIException {
+public class RMAPIResourceNotFoundException extends RMAPIServiceException {
```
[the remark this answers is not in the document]
dmtkachenko: This can be suppressed since there 3 java base classes in the hierarchy

src/main/java/org/folio/rmapi/exception/RMAPIUnAuthorizedException.java
```diff
@@ -4,11 +4,12 @@
  * @author cgodfrey
  *
  */
-public class RMAPIUnAuthorizedException extends RMAPIException {
+public class RMAPIUnAuthorizedException extends RMAPIServiceException {
```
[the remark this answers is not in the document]
dmtkachenko: Same to the above

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
sduvvuri-ebsco: ok, makes sense!

PR #39, opened 2018-11-14
MODKBEKBJ-93: Implement GET /eholdings/root-proxy endpoint

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-93, we are supposed to implement `GET /eholdings/root-proxy` endpoint. This PR addresses that.

## Approach
- Experiment with Ruby code and figure out what request and response look like.
- Put together JSON schemas and RAML for supporting `/eholdings/proxy-types` and `/eholdings/root-proxy`
- Put together root proxy implementation class and associated tests.

## Screenshots
get_root_proxy

### DIFF
--- a/src/main/java/org/folio/rmapi/model/CustomLabel.java
+++ b/src/main/java/org/folio/rmapi/model/CustomLabel.java
@@ -1,0 +1,54 @@
+package org.folio.rmapi.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+public class CustomLabel {
+  
+  @JsonProperty("id")
+  private Integer id;
+  @JsonProperty("displayLabel")
+  private String displayLabel;
+  @JsonProperty("displayOnFullTextFinder")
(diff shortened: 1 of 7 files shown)


src/main/java/org/folio/rmapi/model/CustomLabel.java
```diff
@@ -0,0 +1,54 @@
...
+  @JsonProperty("id")
+  private Integer id;
+  @JsonProperty("displayLabel")
+  private String displayLabel;
+  @JsonProperty("displayOnFullTextFinder")
+  private Boolean displayOnFullTextFinder;
+  @JsonProperty("displayOnPublicationFinder")
+  private Boolean displayOnPublicationFinder;
+  
+  @JsonProperty("id")
```
andrii-paias: I think it's sufficient to only annotate fields with @JsonProperty, no need to annotate getter and setters

PR #40, opened 2018-11-15
MODKBEKBJ-23 - Implement GET /eholdings/packages

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-23 we want to have the ability to get packages by supporting '/eholdings/packages' endpoint
## Approach
* Implemented get packages collection method.
* Mapped packages returned from RM API to a collection of packages
* Added permissions to the request


ramls/traits/filterable.raml
```diff
@@ -9,8 +9,8 @@ queryParameters:
     displayName: Content type
     type: string
     description: |
-      Filter to narrow down results based on content type.
-      Possible values are all, aggregatefulltext, abstractandindex, ebook, ejournal, print, unknown, onlinereference.
+      Filter to narrow down results based on content type. Defaults to all.
+      Possible values are all, aggregatedfulltext, abstractandindex, ebook, ejournal, print, unknown, onlinereference.
     example: ebook
```
carolegodfrey: Good catch on the `aggregatedfulltext` -- wonder if it is possible to define these as an enum (similar to how publicationTypes are handled in titles)

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -49,25 +54,70 @@
...
+    boolean isFilterCustom = Boolean.parseBoolean(filterCustom);
+    Sort nameSort = sort != null ? Sort.valueOf(sort.toUpperCase()) : null;
+    MutableObject<RMAPIService> service = new MutableObject<>();
+    CompletableFuture.completedFuture(null)
+      .thenCompose(o -> configurationService.retrieveConfiguration(new OkapiData(okapiHeaders)))
+      .thenAccept(rmapiConfiguration ->
+        service.setValue(new RMAPIService(rmapiConfiguration.getCustomerId(),
+          rmapiConfiguration.getAPIKey(), rmapiConfiguration.getUrl(), vertxContext.owner())))
+      .thenCompose(o -> service.getValue().getVendors(isFilterCustom))
+      .thenCompose(vendors ->
```
carolegodfrey: Supporting the custom filter adds more complexity -- but logic all seems correct where you need to look up the custom vendor/provider

src/test/java/org/folio/rest/impl/EholdingsPackagesTest.java
```diff
@@ -0,0 +1,129 @@
...
+  @Test
+  public void shouldReturn400WhenCountInvalid() {
+    RestAssured.given()
+      .spec(getRequestSpecification())
+      .when()
+      .get("eholdings/packages?q=American&filter[type]=abstractandindex&count=500")
+      .then()
+      .statusCode(HttpStatus.SC_BAD_REQUEST);
+  }
+
```
carolegodfrey: Could we add a test on the filter for custom packages (since it does the additional rm api vendor look up to support this)

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -49,25 +54,70 @@
...
+  @HandleValidationErrors
+  public void getEholdingsPackages(String filterCustom, String q, String filterSelected,
+    String filterType, String sort, int page, int count, Map<String, String> okapiHeaders,
+    Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+
+    headerValidator.validate(okapiHeaders);
+    packageParametersValidator.validate(filterCustom, filterSelected, filterType, sort);
+
+    boolean isFilterCustom = Boolean.parseBoolean(filterCustom);
+    Sort nameSort = sort != null ? Sort.valueOf(sort.toUpperCase()) : null;
```
andrii-paias: sortable.raml was changed to have "default: relevance", so now sort is guaranteed not to be null

natali-zaitseva: thanks Andrii, will change the code

ramls/traits/filterable.raml
natali-zaitseva: great idea, will do

src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -180,6 +184,36 @@ public RMAPIService(String customerId, String apiKey, String baseURI, Vertx vert
...
+      .build();
+
+    String packagesPath = id == null ? PACKAGES_PATH + "?" : VENDORS_PATH+ '/' + id + '/' + PACKAGES_PATH + "?";
+
+    return this.getRequest(constructURL(packagesPath + path), Packages.class);
+  }
+
+  public CompletableFuture<Vendors> getVendors(boolean filterCustom){
+    CompletableFuture<Vendors> vendorsList = CompletableFuture.completedFuture(new Vendors());
+    if (Boolean.TRUE.equals(filterCustom)) {
```
andrii-paias: This can be simplified to 
if (filterCustom)
similary in PackageParametersValidator condition can be simplified

src/test/java/org/folio/rest/impl/EholdingsPackagesTest.java
```diff
@@ -0,0 +1,129 @@
...
+import org.folio.rest.tools.utils.NetworkUtils;
+import org.folio.rest.util.RestConstants;
+import org.folio.util.TestUtil;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(VertxUnitRunner.class)
+public class EholdingsPackagesTest {
```
andrii-paias: now it's possible to extend WireMockTestBase, to avoid repeating server set up

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
carolegodfrey: I think eventually we may want to cache the customer's vendor id (not something for this PR). Also how we get this will change in future

src/test/java/org/folio/rest/impl/EholdingsPackagesTest.java
natali-zaitseva: added

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -49,25 +54,70 @@
...
-      new HeaderValidator());
+      new HeaderValidator(),
+      new PackageParametersValidator(),
+      new PackagesConverter());
   }
 
   public EholdingsPackagesImpl(RMAPIConfigurationService configurationService,
-                               HeaderValidator headerValidator) {
+                               HeaderValidator headerValidator, PackageParametersValidator packageParametersValidator,
+    PackagesConverter converter) {
```
sduvvuri-ebsco: Spacing here somehow got messed up

src/main/java/org/folio/rest/validator/PackageParametersValidator.java
```diff
@@ -0,0 +1,30 @@
...
+  private static final List<String> FILTER_TYPE_VALUES = Arrays.asList("all", "aggregatedfulltext", "abstractandindex", "ebook", "ejournal", "print", "unknown", "onlinereference");
+
+  public void validate(String filterCustom, String filterSelected, String filterType,
+    String sort) {
+
+    if (!Sort.contains(sort.toUpperCase())){
+      throw new ValidationException("Invalid Query Parameter for sort");
+    }
+    if(Objects.nonNull(filterType) &&  !FILTER_TYPE_VALUES.contains(filterType)){
+      throw new ValidationException("Invalid Query Parameter for filter[:type]");
```
sduvvuri-ebsco: In all these error messages, we do not want the ":" prepended - I think that was unnecessary IN Ruby code too - user is just passing filter[type] - so let's modify it to be consistent - Thanks!

natali-zaitseva: thanks Sobha, I will remove them

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
natali-zaitseva: fixed

PR #45, opened 2018-11-19
MODKBEKBJ-95: GET /eholdings/packages/{packageId} endpoint

andrii-paias:
Jira: MODKBEKBJ-25
Implemented get package by id endpoint.

Refactored exception handling to remove boilerplate code,

## Purpose
Implement getting package by id

## Approach
Reuse mapping from GET /eholdings/packages story and add unit tests


src/main/java/org/folio/rest/util/ErrorHandler.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * ErrorHandler instance can be configured with error mappers for each Exception type by using add* methods,
+ *
+ * When {@link org.folio.rest.util.ErrorHandler#handle(io.vertx.core.Handler, java.lang.Throwable)} method is called,
+ * first mapper that matches exception type is used to construct javax.ws.rs.core.Response and pass it to io.vertx.core.Handler
+ *
+ * Error mappers are checked in the order they were registered
+ */
+public class ErrorHandler {
+
```
carolegodfrey: Awesome -- thanks for doing this and making the error handling in the impl classes so much cleaner!

src/main/java/org/folio/rest/converter/PackagesConverter.java
```diff
@@ -80,4 +101,19 @@ private PackageCollectionItem convertPackage(PackageData packageData) {
       .withRelationships(EMPTY_PACKAGES_RELATIONSHIP);
   }
 
+  private Token convertToToken(TokenInfo packageToken) {
+    if(Objects.isNull(packageToken)){
```
carolegodfrey: Minor comment -- is is possible to have a common method that can be used by both package and vendor/provider?

PR #49, opened 2018-11-20
MODKBEKBJ-53 include packages flag

andrii-paias:
## Purpose
Support '?include=packages' for /eholdings/providers/{provider_id}

## Approach
If include=packages then send request to retrieve packages in parallel

This is a PR only for commit https://github.com/folio-org/mod-kb-ebsco-java/commit/6f06b667495778717a604aa42587ba76e8f40d62
It uses changes from PR https://github.com/folio-org/mod-kb-ebsco-java/pull/45 , 
so it should only be merged after that PR

### DIFF
--- a/ramls/traits/includePackages.raml
+++ b/ramls/traits/includePackages.raml
@@ -3,6 +3,5 @@ queryParameters:
     displayName: Nested resource
     type: string
     description: Name of nested resource to include
-    default: packages
     example: packages
     required: false
--- a/src/main/java/org/folio/rmapi/RMAPIService.java
+++ b/src/main/java/org/folio/rmapi/RMAPIService.java
@@ -1,14 +1,8 @@
 package org.folio.rmapi;
 
-import io.vertx.core.Vertx;
-import io.vertx.core.buffer.Buffer;
-import io.vertx.core.http.HttpClient;
-import io.vertx.core.http.HttpClientRequest;
-import io.vertx.core.http.HttpClientResponse;
-import io.vertx.core.json.Json;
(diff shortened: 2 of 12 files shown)


ramls/traits/includePackages.raml
```diff
@@ -3,6 +3,5 @@ queryParameters:
     displayName: Nested resource
     type: string
     description: Name of nested resource to include
-    default: packages
```
andrii-paias: I removed this "default:", because default should be null in this case, if default is packages then we always get include=packages flag

src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -87,13 +89,13 @@ public RMAPIService(String customerId, String apiKey, String baseURI, Vertx vert
...
         String.format("%s Code = %s Message = %s Body = %s", INVALID_RMAPI_RESPONSE, response.statusCode(),
-            response.statusMessage(), body.toString()),
+          response.statusMessage(), body.toString()),
         response.statusCode(), response.statusMessage(), msgBody, query));
     }
   }
-  
+
   private String mapVendorToProvider(String msgBody) {
-    return msgBody.replaceAll(VENDOR_LOWER_STRING, PROVIDER_LOWER_STRING).replaceAll(VENDOR_UPPER_STRING, PROVIDER_UPPER_STRING);
```
andrii-paias: Here Sonar thinks that using regex is security vulnerability,
changed replaceAll to replace, replace method also replaces all occurences, but it uses fixed string instead of regex

PR #51, opened 2018-11-20
MODKBEKBJ-31: GET resource by resourceId

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-31, we are supposed to implement `GET /eholdings/resources/{resourceId}`. This PR addresses that.

## Approach
- Implement `GET /eholdings/resources/{resourceId}`
- Add unit tests

## Screenshots
get_resource_by_resourceid

### DIFF
--- a/src/main/java/org/folio/rest/converter/CommonAttributesConverter.java
+++ b/src/main/java/org/folio/rest/converter/CommonAttributesConverter.java
@@ -1,12 +1,58 @@
 package org.folio.rest.converter;
 
+import java.util.Objects;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.folio.rest.jaxrs.model.TitleContributors;
+import org.folio.rest.jaxrs.model.TitleIdentifier;
+import org.folio.rest.jaxrs.model.TitleSubject;
+import org.apache.commons.lang3.StringUtils;
+import org.folio.rest.jaxrs.model.Coverage;
+import org.folio.rest.jaxrs.model.PublicationType;
 import org.folio.rest.jaxrs.model.Token;
+import org.folio.rmapi.model.CoverageDates;
+import org.folio.rmapi.model.EmbargoPeriod;
+import org.folio.rmapi.model.Identifier;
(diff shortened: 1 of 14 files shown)


src/main/java/org/folio/rest/converter/CommonAttributesConverter.java
```diff
@@ -0,0 +1,111 @@
...
+    )
+    .collect(Collectors.toList());
+  }
+
+  public org.folio.rest.jaxrs.model.Proxy convertProxy(org.folio.rmapi.model.Proxy proxy) {
+    org.folio.rest.jaxrs.model.Proxy p = new org.folio.rest.jaxrs.model.Proxy();
+    p.setId(proxy.getId());
+    p.setInherited(proxy.getInherited());
+    return p;
+  }
```
carolegodfrey: Great these are common for both resources and titles!

src/main/java/org/folio/rest/impl/EholdingsResourcesImpl.java
```diff
@@ -0,0 +1,154 @@
...
+        RMAPIService rmapiService = new RMAPIService(rmapiConfiguration.getCustomerId(), rmapiConfiguration.getAPIKey(),
+          rmapiConfiguration.getUrl(), vertxContext.owner());
+        return rmapiService.retrieveResource(parsedResourceId);
+    })
+    .thenAccept(title ->
+      asyncResultHandler.handle(Future.succeededFuture(GetEholdingsResourcesByResourceIdResponse
+        .respond200WithApplicationVndApiJson(converter.convertFromRMAPIResource(title)))))
+    .exceptionally(e -> {
+      logger.error(INTERNAL_SERVER_ERROR, e);
+      if (e.getCause() instanceof RMAPIResourceNotFoundException) {
```
carolegodfrey: Eventually this can be updated to use ErrorHandler from - but probably to be coordinated 
 on after https://github.com/folio-org/mod-kb-ebsco-java/pull/45

sduvvuri-ebsco: Yes - agreed! Thanks Carole!

PR #52, opened 2018-11-21
MODKBEKBJ-21 - Implement GET /eholdings/providers/{provider_id}/packages

ivan-brovenko:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-21 we want to have the ability to get packages by provider id by supporting  `/eholdings/providers/{provider_id}/packages` endpoint

## Approach

- Implement get provider packages method
- Added permissions to the request
- Mapped packages returned from RM API by provider id to a collection of packages

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -50,7 +50,8 @@
         },
         {
           "methods": ["GET"],
-          "pathPattern": "/eholdings/providers/{providerId}/packages"
+          "pathPattern": "/eholdings/providers/{providerId}/packages",
+          "permissionRequired": ["kb-ebsco.provider.packages.get"]
         },
         {
           "methods": ["POST"],
@@ -147,6 +148,11 @@
       "description": "Put single provider"
     },
     {
+      "permissionName": "kb-ebsco.provider.packages.get",
+      "displayName": "get packages for a single provider",
+      "description": "get packages for a single provider"
+    },
+    {
       "permissionName": "kb-ebsco.title.get",
       "displayName": "get single title",
       "description": "Get single title"
@@ -197,6 +203,7 @@
         "kb-ebsco.providerCollection.get",
         "kb-ebsco.provider.get",
         "kb-ebsco.provider.put",
+        "kb-ebsco.provider.packages.get",
         "kb-ebsco.titleCollection.get",
         "kb-ebsco.packageCollection.get",
         "kb-ebsco.package.delete",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -51,7 +51,7 @@
         {
           "methods": ["GET"],
           "pathPattern": "/eholdings/providers/{providerId}/packages",
-          "permissionsRequired": ["kb-ebsco.provider.packages.get"]
+          "permissionsRequired": ["kb-ebsco.provider.packageCollection.get"]
         },
         {
           "methods": ["POST"],
@@ -148,7 +148,7 @@
       "description": "Put single provider"
     },
     {
-      "permissionName": "kb-ebsco.provider.packages.get",
+      "permissionName": "kb-ebsco.provider.packageCollection.get",
       "displayName": "get packages for a single provider",
       "description": "get packages for a single provider"
     },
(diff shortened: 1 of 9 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -50,7 +50,8 @@
         },
         {
           "methods": ["GET"],
-          "pathPattern": "/eholdings/providers/{providerId}/packages"
+          "pathPattern": "/eholdings/providers/{providerId}/packages",
+          "permissionRequired": ["kb-ebsco.provider.packages.get"]
```
andrii-paias: There's a typo, it should be "permissionsRequired"

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
```diff
@@ -180,8 +191,45 @@ public void putEholdingsProvidersByProviderId(String providerId, String contentT
...
+  @Validate
   @HandleValidationErrors
-  public void getEholdingsProvidersPackages(String q, String filterSelected, String filterType, String sort, int page, int count, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
-    asyncResultHandler.handle(Future.succeededFuture(GetEholdingsProvidersResponse.status(Response.Status.NOT_IMPLEMENTED).build()));
+  public void getEholdingsProvidersPackagesByProviderId(String providerId, String q, String filterSelected,
+                                                        String filterType, String sort, int page, int count,
+                                                        Map<String, String> okapiHeaders,
+                                                        Handler<AsyncResult<Response>> asyncResultHandler,
+                                                        Context vertxContext) {
+    long providerIdLong;
```
andrii-paias: In all other cases we validate that q is not empty and return an error if it's empty, I think it's better to do it here too, for consistency

ramls/examples/providers/providers_providerId_packages_get_400_response.json
```diff
@@ -0,0 +1,10 @@
+{
+  "errors": [
+    {
+      "title": "Parameter OrderBy should be by PackageName when parameter Search is empty."
```
andrii-paias: error message "Parameter OrderBy should be by PackageName when parameter Search is empty" is an error returned from RM API, mod-kb-ebsco-java doesn't have parameter OrderBy, in other endpoints we return "Search parameter cannot be empty" for this case

descriptors/ModuleDescriptor-template.json
```diff
@@ -197,6 +203,7 @@
         "kb-ebsco.providerCollection.get",
         "kb-ebsco.provider.get",
         "kb-ebsco.provider.put",
+        "kb-ebsco.provider.packages.get",
```
carolegodfrey: This is very minor -- but just wonder if we should be calling this permission
`kb-ebsco.provider.packageCollection.get`?  or if `kb-ebsco.provider.packages.get` is more readable

ramls/providers.raml
```diff
@@ -88,20 +89,36 @@ traits:
...
+    /packages:
+        get:
+          description: Search within a list of packages associated with a given provider.
+          is: [queriable, filterable, sortable, pageable]
+          responses:
+            200:
+              description: OK
+              body:
+                application/vnd.api+json:
+                  description: List of providers matching the provider name and the total number of providers found.
```
carolegodfrey: Minor adjustment - Should this description be something more like
`List of packages associated with a given provider matching search criteria`

ivan-brovenko: Thanks, updated

descriptors/ModuleDescriptor-template.json
ivan-brovenko: Thanks, updated

ramls/examples/providers/providers_providerId_packages_get_400_response.json
ivan-brovenko: Thanks, updated

src/main/java/org/folio/rest/impl/EholdingsProvidersImpl.java
ivan-brovenko: Thanks, updated

descriptors/ModuleDescriptor-template.json
ivan-brovenko: Thanks, updated

PR #53, opened 2018-11-26
Implement PUT /eholdings/packages/{packageId} endpoint

andrii-paias:
Jira: MODKBEKBJ-26

## Purpose
Implement "update package by id" endpoint.

## Approach
Used module mod-kb-ebsco as reference

#### TODOS and Open Questions
There is an issue with attributes that can't be updated:
Currently UI sends all package attributes in PUT request, 
so using schema packagePutDataAttributes.json results in error, because UI sends some additional unexpected attributes.

In this PR I followed approach used in PUT provider endpoint and changed packagePutDataAttributes.json schema to packageDataAttributes.json, 
so that all attributes are present in schema, but only some of them are actually used for update.

Another approach I see is to use packagePutDataAttributes.json with "additionalProperties": true, that way all unused attributes will be mapped to a separate map "Map<String, Object> additionalProperties"

- [ ] Another small issue: ruby module responds with error "Cannot set allowEbscoToAddTitles for custom package" on update request for custom package if request has "allowKbToAddTitles": true, 
But it ignores all other fields that are not updatable, and only updates fields that are allowed to be updated. I changed it to ignore "allowKbToAddTitles" for custom packages as well.


ramls/types/packages/packagePutData.json
```diff
@@ -15,7 +21,7 @@
     "attributes": {
       "description": "Data type attributes",
       "type": "object",
-      "$ref": "packagePutDataAttributes.json"
+      "$ref": "packageDataAttributes.json"
```
carolegodfrey: Interesting other approach that you mentioned for this to `keep packagePutDataAttributes.json with "additionalProperties": true` -- we would explicitly only define the fields that are updated in that case - but this is probably fine and consistent with Provider

src/main/java/org/folio/rest/validator/PackagePutBodyValidator.java
```diff
@@ -0,0 +1,51 @@
...
+      .map(Token::getValue)
+      .orElse(null);
+
+    if (isSelected == null || !isSelected) {
+      ValidatorUtil.checkFalseOrNull("allowKbToAddTitles", allowKbToAddTitles);
+      ValidatorUtil.checkFalseOrNull("visibilityData.isHidden", isHidden);
+      ValidatorUtil.checkIsEmpty("customCoverage.beginCoverage", beginCoverage);
+      ValidatorUtil.checkIsEmpty("customCoverage.endCoverage", endCoverage);
+      ValidatorUtil.checkIsNull("packageToken.value", value);
+    }
```
carolegodfrey: Looks like a good translation for ruby code which checks for selected to be true  https://github.com/folio-org/mod-kb-ebsco/blob/9b68934c587983e28cb7a3d9c73e7a562b31ac90/app/controllers/validation/package_parameters.rb#L16

src/main/java/org/folio/rest/validator/CustomPackagePutBodyValidator.java
```diff
@@ -0,0 +1,37 @@
...
+    String beginCoverage = Optional.ofNullable(attributes.getCustomCoverage())
+      .map(Coverage::getBeginCoverage)
+      .orElse(null);
+
+    String endCoverage = Optional.ofNullable(attributes.getCustomCoverage())
+      .map(Coverage::getEndCoverage)
+      .orElse(null);
+
+    ValidatorUtil.checkIsNotEmpty("name", name);
+    ValidatorUtil.checkIsNotNull("contentType", contentType);
```
carolegodfrey: Do we also want to check the validity of the content type? We may not have been doing this in the ruby code per below --but there was a note for checking it

https://github.com/folio-org/mod-kb-ebsco/blob/9b68934c587983e28cb7a3d9c73e7a562b31ac90/app/controllers/validation/custom_package_parameters.rb#L36

src/test/java/org/folio/rest/impl/EholdingsPackagesTest.java
```diff
@@ -189,28 +190,195 @@ public void shouldReturn400WhenPackageIdIsInvalid() {
...
+    RestAssured.given()
+      .spec(getRequestSpecification())
+      .header(CONTENT_TYPE_HEADER)
+      .when()
+      .body(TestUtil.readFile("requests/kb-ebsco/package/put-package-not-selected-non-empty-fields.json"))
+      .put("eholdings/packages/" + providerId + "-" + packageId)
+      .then()
+      .statusCode(HttpStatus.SC_UNPROCESSABLE_ENTITY);
+  }
+
```
carolegodfrey: Just curious if a 400 test may be need for coverage as well?

src/main/java/org/folio/rest/validator/CustomPackagePutBodyValidator.java
andrii-paias: ContentType is enum, so this validation is automatically handled by RMB module

carolegodfrey: Ok thanks for clarification @andrii-paias

PR #54, opened 2018-11-26
MODKBEKBJ-24 - Implement POST /eholdings/packages

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-24 we want to have the ability to post packages by supporting '/eholdings/packages' endpoint

## Approach
Implemented get packages collection method.
Mapped packages returned from RM API to a collection of packages
Added permissions to the request

#### TODOS and Open Questions
- [ ] there is a question about Content-type. In Ruby code we use 'application/json' but currently for Post request RMB supports following content types 

image and the is two ways - create PR to RMB for adding our content Type or change it in our raml file.


ramls/packages.raml
```diff
@@ -66,7 +66,7 @@ types:
       Content-Type:
         example: application/vnd.api+json
     body:
-      application/vnd.api+json:
+      application/json:
```
sduvvuri-ebsco: As discussed, let's please wait on this until we hear back from RMB folks! Thanks!

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -120,8 +127,36 @@ public void getEholdingsPackages(String filterCustom, String q, String filterSel
...
+
+    PackagePost packagePost = converter.convertToPackage(entity);
+
+    MutableObject<RMAPIService> service = new MutableObject<>();
+    CompletableFuture.completedFuture(null)
+      .thenCompose(o -> configurationService.retrieveConfiguration(new OkapiData(okapiHeaders)))
+      .thenAccept(rmapiConfiguration ->
+        service.setValue(new RMAPIService(rmapiConfiguration.getCustomerId(),
+          rmapiConfiguration.getAPIKey(), rmapiConfiguration.getUrl(), vertxContext.owner())))
+      .thenCompose(vendors  ->  service.getValue().postPackage(packagePost))
```
sduvvuri-ebsco: shall we rename 'vendors' to 'package' instead?

natali-zaitseva:
should be merged after [https://github.com/folio-org/raml-module-builder/pull/334/files](url)

natali-zaitseva:
should be merged after #https://github.com/folio-org/raml-module-builder/pull/334

src/test/resources/responses/rmapi/packages/get-created-package-response.json
```diff
@@ -0,0 +1,52 @@
+{
```
andrii-paias: this file and post-package-request.json probably shouldn't be in  rmapi/packages/ subfolder, because they are request/response for kb-ebsco-java module

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
natali-zaitseva: 'package' is one of the Java Language Keywords and variable can not have the same name. I have renamed it to 'packageById'

src/test/resources/responses/rmapi/packages/get-created-package-response.json
natali-zaitseva: moved to other package

ramls/packages.raml
carolegodfrey: Looks like Adam has merged related PR in RMB project

natali-zaitseva: updated the code and changed the body type to application/vnd.api+json:

PR #56, opened 2018-11-28
MODKBEKBJ-73, MODKBEKBJ-74 and MODKBEKBJ-75: Support '?include=title' and '?include=provider' and '?include=package' for /eholdings/resources/{resourceId}

andrii-paias:
Jira: MODKBEKBJ-75

## Purpose
Support '?include=title' flag for /eholdings/resources/{resourceId} endpoint
Support '?include=provider' flag for /eholdings/resources/{resourceId} endpoint
Support '?include=package' flag for /eholdings/resources/{resourceId} endpoint

## Approach
Used data that is already present in resource to populate title,
Retrieved provider by id to include it in response,
moved some test fixtures to json files, and used JsonAssert to
compare expected response with actual response,

Changed metaIncluded.json type to hasOneRelationship.json type,
so that relationship can contain either "meta" or "data" attribute

Changed "included" type from resourceIncludedItem.json to java.lang.Object,
so that classes from existing mappings can be added as "included"

#### TODOS and Open Questions
- [ ] Currently Ruby module return an array of objects in "included" attribute, 
objects in array can have structure of title,provider,package or some combination of those. 
In this PR I modeled this with array that has "javaType" : "java.lang.Object". So any object can be added to this array. If this is too general, then it's possible to make package,title,provider classes to implement a marker interface, for example "IncludableItem", and change included array to only contain IncludableItem objects


src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -341,4 +355,22 @@ public Packages getPackages() {
       return packages;
     }
   }
+
+  public static class ResourceResult {
```
dmtkachenko: I don't see why it has to be an inner class. To me it looks like a part of RM API Service model, like Title VendorId. Otherwise the methods look inconsistent, in some cases you return model classes in other - something else which belongs to Service

andrii-paias: Other model classes serve as mappings for RM API responses, ResourceResult and VendorResult are just wrapper classes that combine several RM API responses. Agree that they shouldn't be declared as inner classes, moved them to rmapi/result package

ramls/types/resources/resource.json
```diff
@@ -16,7 +16,7 @@
       "description": "List of included items",
       "items": {
         "type": "object",
-        "$ref": "resourceIncludedItem.json"
+        "javaType" : "java.lang.Object"
```
sduvvuri-ebsco: A little confused by this change..resourceIncludedItem.json has all possible fields that can be included for a provider, package or title. Why could we just not use that?

ramls/types/resources/resourceRelationships.json
```diff
@@ -9,17 +9,17 @@
     "provider": {
       "type": "object",
       "description": "Relationship of provider for a given object",
-      "$ref": "../metaIncluded.json"
+      "$ref": "../hasOneRelationship.json"
```
sduvvuri-ebsco: This looks good and is the right thing to do.

src/main/java/org/folio/rest/converter/ResourcesConverter.java
```diff
@@ -69,10 +75,53 @@ public ResourcesConverter(CommonAttributesConverter commonConverter) {
...
-          .withRelationships(EMPTY_RELATIONSHIPS)
+          .withRelationships(createEmptyRelationship())
         )
         .withIncluded(null)
-        .withJsonapi(RestConstants.JSONAPI))
-      .collect(Collectors.toList());
+        .withJsonapi(RestConstants.JSONAPI);
+      resultResource.setIncluded(new ArrayList<>());
+      if (includeTitle) {
+        resultResource.getIncluded().add(titleConverter.convertFromRMAPITitle(title, null).getData());
```
sduvvuri-ebsco: This comment is not related to this PR but it will be good to have consistency in naming converter methods -> convertFromRMAPITitle, convertToProvider, convert etc.

PR #61, opened 2018-12-03
MODKBEKBJ-28 -  Implement GET /eholdings/packages/{packageId}/resources

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-28 - implement GET /eholdings/packages/{packageId}/resources.

Method parameters include
- filter[selected]
- filter[type]
- filter[name]
- filter[isxn]
- filter[subject]
- filter[publisher]
- sort
- page
- count

Some sample GET requests to be supported include:

- /eholdings/packages/{packageId}/resources
- /eholdings/packages/{packageId}/resources?page=2
- /eholdings/packages/{packageId}/resources?filter[name]=test
- /eholdings/packages/{packageId}/resources?filter[publisher]=test
- /eholdings/packages/{packageId}/resources?filter[subject]=test
- /eholdings/packages/{packageId}/resources?filter[isxn]=1362-3613
- /eholdings/packages/{packageId}/resources?filter[selected]=ebsco
- /eholdings/packages/{packageId}/resources?filter[type]=book
-  /eholdings/packages/{packageId}/resources?filter[type]=book&sort=relevance&count=10

## Approach
- Added method permissions
- Adjusted RAML to include page and count, pageable trait
- Re-Used titleParametersValidator class to validate incoming parameters. Needed to account for default case of ` /eholdings/packages/{packageId}/resources` which doesn't have default filter.
- Added method to resourcesConverter to support resourceCollection
- Added null checks to commonAttributesConverter methods 
- updated queriableUrlBuilder to add default `search=`
- updated TitlesFilterableUrlBuilderto default `searchfield=titlename`
- Moved getResponseWithStatus method to WireMockTestBase class since used in multiple classes
- Added test to confirm resource response and confirm correct request is being constructed to RM API
- Updated deserialize of Error response from RM API. In the case of 400 errors returned for title requests -- we are seeing an error response body that differs in case from other endpoint error responses.

For example:

```
{
  "Errors": [
    {
      "Code": 1006,
      "Message": "Parameter Count is outside the range 1-100.",
      "SubCode": 0
    }
  ]
}
```

verses
```
{
  "errors": [
    {
      "code": 1009,
      "subCode": 0,
      "message": "Custom Package with the provided name already exists"
    }
  ]
}
```

#### TODOS and Open Questions
- [ ] Added  @SuppressWarnings("squid:S00107") to EholdingsPackagesImpl constructor. It now takes 9 parameters due to the varied validators and converters. Another option would be to create
TitleParametersValidator and ResourcesConverter inside method getEholdingsPackagesResourcesByPackageId but this is inconsistent with others.
- [x] Issue handling RMAPIServiceException
- [x] Confirm Count > 100 is be handled by RMB framework - needed annotations


src/test/java/org/folio/rest/impl/EHoldingsProxyTypesImplTest.java
```diff
@@ -57,14 +57,6 @@ public void shouldReturnForbiddenWhenRMAPIRequestCompletesWith403ErrorStatus() {
     getResponse(EHOLDINGS_PROXY_TYPES_URL).statusCode(HttpStatus.SC_FORBIDDEN);
   }
 
-  private ExtractableResponse<Response> getResponseWithStatus(String resourcePath, int expectedStatus) {
```
sduvvuri-ebsco: Why is this test removed as part of this PR?

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
```diff
@@ -462,15 +462,6 @@ private Provider getUpdatedProvider() {
               .withIncluded(false)))));
   }
 
-  private ExtractableResponse<Response> getResponseWithStatus(String resourcePath, int expectedStatus) {
```
sduvvuri-ebsco: same question as above

src/test/java/org/folio/rest/impl/EholdingsTitlesTest.java
```diff
@@ -215,13 +229,4 @@ public void shouldReturnTitleWithoutResourcesWhenInvalidInclude() throws IOExcep
 
     JSONAssert.assertEquals(expected, actual, false);
   }
-
-  private ExtractableResponse<Response> getResponseWithStatus(String resourcePath, int expectedStatus) {
```
sduvvuri-ebsco: ""

sduvvuri-ebsco: ok, never mind, I got it. Its been moved to WireMockTestBase.

src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
carolegodfrey: Moved the code to base class -- WireMockTestBase.java since it was used in multiple places

src/test/java/org/folio/rest/impl/EholdingsTitlesTest.java
carolegodfrey: As above -- in base class

src/test/java/org/folio/rest/impl/EHoldingsProxyTypesImplTest.java
carolegodfrey: Moved method to base class

src/test/java/org/folio/rest/impl/EholdingsPackagesTest.java
```diff
@@ -477,6 +468,171 @@ public void shouldReturn400WhenPackagePostDataIsInvalid()throws URISyntaxExcepti
       .statusCode(HttpStatus.SC_BAD_REQUEST);
   }
 
+  @Test
+  public void shouldReturnDefaultResourcesOnGetWithResources() throws IOException, URISyntaxException {
```
andrii-paias: I feel like those tests only test conversion of mod-kb-ebsco url into RM API url,
since this conversion is only done in TitlesFilterableUrlBuilder, I think it's better to test those different cases directly in TitlesFilterableUrlBuilderTest, that way tests will be faster and will only break if url mapping breaks,
And then it's good to leave 1-2 wiremock tests to check that everything is wired correctly

src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -231,13 +231,21 @@ private void addRequestHeaders(HttpClientRequest request) {
   }
 
   public CompletableFuture<Titles> retrieveTitles(FilterQuery filterQuery, Sort sort, int page, int count) {
+
```
andrii-paias: Small remark, I think it's ok to have two different methods like this:
```
  public CompletableFuture<Titles> retrieveTitles(FilterQuery filterQuery, Sort sort, int page, int count) {
    String path = new TitlesFilterableUrlBuilder()
      .filter(filterQuery)
      .sort(sort)
      .page(page)
      .count(count)
      .build();
    return this.getRequest(constructURL(TITLES_PATH + "?" + path), Titles.class);
  }

  public CompletableFuture<Titles> retrieveTitles(Long providerId, Long packageId, FilterQuery filterQuery, Sort sort, int page, int count) {
    String path = new TitlesFilterableUrlBuilder()
      .filter(filterQuery)
      .sort(sort)
      .page(page)
      .count(count)
      .build();

    return this.getRequest(constructURL(
      (VENDORS_PATH + '/' + providerId + '/' + PACKAGES_PATH + '/' + packageId + '/' + TITLES_PATH + "?") + path), Titles.class);
  }
```

That way it won't be necessary to pass nulls, and to check for nulls, also those methods don't really have code duplication because they just pass parameters into TitlesFilterableUrlBuilder and into getRequest

src/main/java/org/folio/rest/converter/CommonAttributesConverter.java
```diff
@@ -65,6 +61,7 @@ public Token convertToken(TokenInfo tokenInfo) {
   }
 
   public List<TitleContributors> convertContributors(List<org.folio.rmapi.model.Contributor> contributorList) {
+    if(Objects.isNull(contributorList)) return Collections.emptyList();
```
andrii-paias: According to Java convention every if statement should use curly braces to avoid confusion about what lines will be executed

src/main/java/org/folio/rmapi/RMAPIService.java
carolegodfrey: Good point -- I originally started off with 2 separate methods and then consolidated. But may make it easier in long run to keep as separate -- will adjust

src/main/java/org/folio/rest/converter/CommonAttributesConverter.java
carolegodfrey: Ok, will adjust -- think the single line shorthand/ single line was suggested through IDE plugin (just recently switched to Intellij)

src/test/java/org/folio/rest/impl/EholdingsPackagesTest.java
carolegodfrey: Ok, left 2 tests to confirm overall wiring to expected RM API response - moved other tests to TitlesFilterableUrlBuilderTest that were not redundant with any existing

PR #62, opened 2018-12-03
MODKBEKBJ-71 - API Tests DELETE /eholdings/packages/{packageId} endpoint

ivan-brovenko:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-65 we want to have API tests for POST /eholdings/packages endpoint and for api tests we need to swap error messages for title and details

## Approach

- swapped error.title with error.details messages

### DIFF
--- a/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
+++ b/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
@@ -100,12 +100,12 @@ public class EholdingsPackagesImpl implements EholdingsPackages {
   @Validate
   @HandleValidationErrors
   public void getEholdingsPackages(String filterCustom, String q, String filterSelected,
-    String filterType, String sort, int page, int count, Map<String, String> okapiHeaders,
-    Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -247,9 +247,9 @@ public void deleteEholdingsPackagesByPackageId(String packageId, Map<String, Str
           rmapiConfiguration.getUrl(), vertxContext.owner()));
         return rmapiService.getValue().retrievePackage(parsedPackageId);
       })
-      .thenCompose( packageData -> {
-        if(!packageData.getIsCustom()){
-          throw new InputValidationException(INVALID_PACKAGE_TITLE, INVALID_PACKAGE_DETAILS);
+      .thenCompose(packageData -> {
+        if (!packageData.getIsCustom()) {
+          throw new InputValidationException(INVALID_PACKAGE_DETAILS, INVALID_PACKAGE_TITLE);
```
sduvvuri-ebsco: Instead of swapping the variables, can we just swap error messages instead, the place they are assigned to these variables? Thanks!

ivan-brovenko: Thanks, updated

PR #70, opened 2018-12-10
MODKBEKBJ-114: Change cache implementation to use vertx.sharedData

andrii-paias:
## Purpose
Use the way to share data between verticles that is provided by Vert.x instead of relying on singleton with static variable
## Approach
Made RMAPIConfiguration immutable and implemented cache by using
vertx SharedData object

### DIFF
--- a/src/main/java/org/folio/config/RMAPIConfiguration.java
+++ b/src/main/java/org/folio/config/RMAPIConfiguration.java
@@ -30,10 +30,4 @@ public final class RMAPIConfiguration implements Shareable {
   public Boolean getConfigValid() {
     return configValid;
   }
-
-  @Override
-  public String toString() {
-    return "RMAPIConfiguration [customerId=" + customerId
-      + ", apiKey=" + apiKey + ", url=" + url + ']';
-  }
 }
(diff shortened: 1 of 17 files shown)


src/main/java/org/folio/config/RMAPIConfiguration.java
```diff
@@ -25,22 +31,6 @@ public Boolean getConfigValid() {
...
-  public void setUrl(String url) {
-    this.url = url;
-  }
-
-  public void setConfigValid(Boolean configValid) {
-    this.configValid = configValid;
-  }
-
   @Override
   public String toString() {
```
dmtkachenko: you can potentially remove toString(), it can be generated by Lombook for @Value classes

andrii-paias: Thanks, done

src/main/java/org/folio/config/cache/RMAPIConfigurationCache.java
```diff
@@ -1,36 +1,58 @@
...
-    cache.invalidate(tenant);
+    getLocalMap().remove(tenant);
+  }
+
+  private LocalMap<String, RMAPIConfigurationWrapper> getLocalMap() {
+    return vertx.sharedData().getLocalMap("configurationMap");
+  }
+
+  @Value
+  @Builder(toBuilder = true)
```
dmtkachenko: Why do you need a builder if you use a constructor?

PR #71, opened 2018-12-11
Modkbekbj 106 add spring di

andrii-paias:
## Purpose
Use Spring for dependency injection instead of creating objects when they are used

## Approach
Dependency injection with spring is implemented in the following way:
Spring context is created in InitAPIImpl and added to Vertx context.
Spring context uses configuration ApplicationConfig, this configuration
uses @ComponentScan annotation to find and instantiate classes annotated
with @Component. It also declares PropertySourcesPlaceholderConfigurer that reads
application.properties file and uses its properties to populate beans.
Spring doesn't control api classes(e.g. EholdingsPackagesImpl) so beans
from context are autowired to api classes by using SpringContextUtil.autowireDependencies
method. This method gets Spring context from Vertx context and uses it
to populate api class.

More details can be found on wiki: [Spring DI in Vertx overview](https://wiki.folio.org/display/FOLIJET/Spring+DI+in+Vertx+overview+-+DRAFT)

## Learning
There are several libraries that integrate Vertx with Spring:
https://github.com/amoAHCP/spring-vertx-ext
https://github.com/rworsnop/vertx-beans
https://github.com/juanavelez/vertx-spring-verticle-factory

Most of the libraries integrate with Spring context by requiring user to add a custom verticle, since we already use RestVerticle and we can't replace it, we can't use those libraries.

I see two approaches for adding Spring context to Vertx:
1)Use Vertx context to store Spring context. With this approach each verticle has a separate Spring context, this creates some memory overhead because context is constructed multiple times, but it preserves the guarantee of thread-safety, because verticles don't share objects.
I used this approach. Also we currently only create 1 verticle per module (if module is deployed through fat jar, it will always default to having 1 verticle, and I don't see a way to customize amout of verticles when running jar file)
2) Second approach is to share Spring context by storing it in a static field (or in Vertx LocalMap), that way all verticles will have access to the same objects. This will allow to, for example, share cache between verticles. 
But Vertx discourages sharing of mutable objects because it removes thread safety guarantees and requires blocking of event loop threads.


src/main/java/org/folio/rest/impl/EHoldingsRootProxyImpl.java
```diff
@@ -30,35 +28,27 @@
...
-
-  public EHoldingsRootProxyImpl(RMAPIConfigurationService configurationService,
-                                HeaderValidator headerValidator,
-                                RootProxyConverter converter,
-                                RootProxyPutBodyValidator bodyValidator) {
-    this.configurationService = configurationService;
-    this.headerValidator = headerValidator;
-    this.converter = converter;
-    this.bodyValidator = bodyValidator;
+  @SuppressWarnings("squid:S1172")
```
carolegodfrey: Will this be required now for all constructors due to di?

andrii-paias: Yes, this is not related to di, Sonar generates a warning here because tenanId isn't used, but in reality RestVerticle uses reflection to find a constructor with Vertx and String parameters. So without tenantId parameter RestVerticle won't call this constructor.
Here's code from RestVerticle.
```
try {
    o = aClass.getConstructor(Vertx.class, String.class).newInstance(this.vertx, tenantId[0]);
} catch (Exception var36) {
    o = aClass.newInstance();
}
```

dmtkachenko: A little research of Vertx code has shown that we can use default constructor and obtain Vertx instance from the current context:
- `VertxUtils.getVertxFromContextOrNew()` or
- 'Vertx.currentContext().owner()'

2nd case is more strict and states that we don't create a new Vertx instance since it should be already available

PR #74, opened 2018-12-11
MODKBEKBJ-32: Implement PUT for custom and managed resources

sduvvuri-ebsco:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-32, we are supposed to implement PUT for managed as well as custom resources. This PR addresses that.

## Approach
- Update module-descriptor to include this method and related permissions
- Support update to managed and custom resources
- Managed resources can be updated only if selected; else error
- Add validations for fields that can be updated in both managed and custom packages
- Add relevant unit tests

#### TODOS and Open Questions
- [ ] Create a gif for this PR; will do tomorrow

### DIFF
--- a/ramls/types/identifier.json
+++ b/ramls/types/identifier.json
@@ -14,13 +14,13 @@
     "subtype": {
       "type": "string",
       "description": "Subtype for the identifier. Valid values are  Print, Online.",
-      "enum": [ "Print", "Online"],
+      "enum": [ "Empty", "Print", "Online", "Preceding", "Succeeding", "Regional", "Linking", "Invalid"],
       "example": "Print"
     },
     "type": {
       "type": "string",
       "description": "Type of identifier. Valid values are ISSN, ISBN.",
-      "enum": [ "ISSN", "ISBN"],
+      "enum": [ "ISSN", "ISBN", "TSDID", "SPID", "EjsJournalID", "NewsbankID", "ZDBID", "EPBookID", "Mid", "BHM"],
       "example": "ISSN"
     }
   }
--- a/src/main/java/org/folio/rest/impl/EholdingsResourcesImpl.java
+++ b/src/main/java/org/folio/rest/impl/EholdingsResourcesImpl.java
@@ -39,7 +39,8 @@ import org.folio.rmapi.model.PackageByIdData;
 import org.folio.rmapi.model.ResourceSelectedPayload;
 import org.folio.rmapi.model.Title;
 import org.folio.rmapi.result.ObjectsForPostResourceResult;
-
+import org.folio.rest.validator.ResourcePutBodyValidator;
+import org.folio.rmapi.model.ResourcePut;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
@@ -51,6 +52,7 @@ import io.vertx.core.logging.LoggerFactory;
 public class EholdingsResourcesImpl implements EholdingsResources{
   private static final String RESOURCE_NOT_FOUND_MESSAGE = "Resource not found";
   private static final int MAX_TITLE_COUNT = 100;
+  private static final String PUT_RESOURCE_ERROR_MESSAGE = "Failed to update package";
 
   private final Logger logger = LoggerFactory.getLogger(EholdingsResourcesImpl.class);
 
@@ -59,7 +61,8 @@ public class EholdingsResourcesImpl implements EholdingsResources{
   private ResourcesConverter converter;
   private IdParser idParser;
   private ResourcePostValidator postValidator;
-
+  private ResourcePutBodyValidator resourcePutBodyValidator;
+  
   public EholdingsResourcesImpl() {
     this(
       new RMAPIConfigurationServiceCache(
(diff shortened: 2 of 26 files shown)


src/main/java/org/folio/rest/impl/EholdingsResourcesImpl.java
```diff
@@ -164,7 +170,45 @@ public void getEholdingsResourcesByResourceId(String resourceId, String include,
...
+    CompletableFuture.completedFuture(null)
+      .thenCompose(o -> configurationService.retrieveConfiguration(new OkapiData(okapiHeaders), vertxContext))
+      .thenCompose(rmapiConfiguration -> {
+        rmapiService.setValue(new RMAPIService(rmapiConfiguration.getCustomerId(),
+          rmapiConfiguration.getAPIKey(), rmapiConfiguration.getUrl(), vertxContext.owner()));
+        return rmapiService.getValue().retrieveResource(parsedResourceId, null);
+      })
+      .thenCompose(resourceData -> {
+        ResourcePut resourcePutBody;
+        boolean isTitleCustom = resourceData.getTitle().getIsTitleCustom().booleanValue();
```
andrii-paias: .booleanValue() is unnecessary here and in one other place, Boolean is unboxed to boolean automatically

src/main/java/org/folio/rest/validator/ResourcePutBodyValidator.java
```diff
@@ -0,0 +1,91 @@
...
+          ValidatorUtil.checkMaxLength("edition", edition, 250);
+        }
+        if(!StringUtils.isBlank(description)) {
+          ValidatorUtil.checkMaxLength("description", description, 1500);
+        }
+        if(!StringUtils.isBlank(url)) {
+          ValidatorUtil.checkMaxLength("url", url, 600);
+          ValidatorUtil.checkUrlFormat("url", url);
+        }
+        if(Objects.nonNull(attributes.getIdentifiers())) {
```
andrii-paias: I just checked that if array is null in json then it defaults to empty list, so there's no need to check for null here and in a few places below:
Objects.nonNull(attributes.getCustomCoverages()
!Objects.isNull(customCoverages)

src/main/java/org/folio/rest/validator/ResourcePutBodyValidator.java
```diff
@@ -0,0 +1,91 @@
...
+        attributes.getCustomCoverages().forEach(customCoverage -> {
+          ValidatorUtil.checkDateValid("beginCoverage", customCoverage.getBeginCoverage());
+          ValidatorUtil.checkDateValid("endCoverage", customCoverage.getEndCoverage());
+        });  
+      }
+    } else {
+      Boolean isHidden = attributes.getVisibilityData() != null ? attributes.getVisibilityData().getIsHidden() : null;
+      EmbargoUnit embargoUnit = attributes.getCustomEmbargoPeriod() != null ? attributes.getCustomEmbargoPeriod().getEmbargoUnit() : null;
+      List<Coverage> customCoverages = attributes.getCustomCoverages();
+      if (!isTitleCustom && (!Objects.isNull(cvgStmt) || !Objects.isNull(embargoUnit) || isHidden || (!Objects.isNull(customCoverages) && !customCoverages.isEmpty()))) {
```
andrii-paias: if isHidden will be null then this will cause NPE, it should be checked for null, or default value should be changed from null to false

src/main/java/org/folio/rest/validator/ValidatorUtil.java
```diff
@@ -111,4 +129,18 @@ public static boolean isUrlValid(String url) {
       return false;
     }
   }
+
+  public static void checkUrlFormat(String paramName, String value) {
+    if(!StringUtils.startsWith(value, "http://") && !StringUtils.startsWith(value, "https://")) {
```
andrii-paias: There is already isUrlValid method, so this can be refactored to "if(!isUrlValid(value))"

src/main/java/org/folio/rest/validator/ValidatorUtil.java
```diff
@@ -111,4 +129,18 @@ public static boolean isUrlValid(String url) {
...
+      throw new InputValidationException(
+          String.format(INVALID_FIELD_FORMAT, paramName),
+          String.format(MUST_BE_VALID_URL, paramName));
+    }
+  }
+
+  public static void checkIdentifierValid(String paramName, Identifier identifier) {
+    if(Objects.isNull(identifier.getId()) || identifier.getId().length() > 20 ){
+      throw new InputValidationException("INVALID_FIELD_FORMAT", paramName);
```
andrii-paias: This will return {"title": "INVALID_FIELD_FORMAT",  "details":"identifier"}

if we need more descriptive error messages then this will work:
checkIsNotNull(paramName, identifier);
checkMaxLength(paramName, identifier.getId(), 20);

src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -446,4 +445,27 @@ private String constructURL(String path) {
     LOG.info("constructurl - path=" + fullPath);
     return fullPath;
   }
+
+  public static class VendorResult {
```
andrii-paias: I guess this class was merged accidentally, on master I moved VendorResult to it's own file, so this can be deleted.

sduvvuri-ebsco: Thanks for all your feedback @andrii-paias ..addressed 👍

ramls/types/identifier.json
```diff
@@ -14,13 +14,13 @@
     "subtype": {
       "type": "string",
       "description": "Subtype for the identifier. Valid values are  Print, Online.",
-      "enum": [ "Print", "Online"],
+      "enum": [ "Empty", "Print", "Online", "Preceding", "Succeeding", "Regional", "Linking", "Invalid"],
```
carolegodfrey: Do we need the other types if we are only concerned with Print and Online and ISSN and ISBN?

sduvvuri-ebsco: I felt like we should not need them but when I copied a payload from UI and used that for PUT, then our code was blowing up. But, now that I think about it, the reason the payload in the UI contained other types is because Ruby module is not filtering those in the GET response for resourceId where as we filter it in Java module now. So, probably, there should not be a situation where a PUT from UI using Java code should supply other types other than Print/Online. I will remove these and hopefully it should be fine.

PR #75, opened 2018-12-12
MODKBEKBJ-62: Support '?include=resources' for /eholdings/packages/{packageId}

andrii-paias:
## Purpose
Support '?include=resources' for /eholdings/packages/{packageId}

## Approach
Retrieved resources by package id and added them to package response if
include flag is set to include=resources.
Moved package test fixture to json file, made other small test refactorings

### DIFF
--- a/src/main/java/org/folio/rmapi/RMAPIService.java
+++ b/src/main/java/org/folio/rmapi/RMAPIService.java
@@ -40,6 +40,7 @@ import org.folio.rmapi.model.Vendor;
 import org.folio.rmapi.model.VendorById;
 import org.folio.rmapi.model.VendorPut;
 import org.folio.rmapi.model.Vendors;
+import org.folio.rmapi.result.PackageResult;
 import org.folio.rmapi.result.ResourceResult;
 import org.folio.rmapi.result.VendorResult;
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/rmapi/RMAPIService.java
```diff
@@ -325,6 +327,21 @@ public Long getFirstProviderElement(Vendors vendors) {
       .thenCompose(vendorResult -> CompletableFuture.completedFuture(vendorResult.getVendor()));
   }
 
+  public CompletableFuture<PackageResult> retrievePackage(PackageId packageId, List<String> includedObjects) {
+    CompletableFuture<PackageByIdData> packageFuture = retrievePackage(packageId);
+    CompletableFuture<Titles> titlesFuture;
+    if (includedObjects.contains(INCLUDE_RESOURCES_VALUE)) {
```
sduvvuri-ebsco: is it a possibility that this "includedObjects" can be null?

andrii-paias: no, if include flag is null then "includedObjects" will be set to empty list

sduvvuri-ebsco: Sounds good!

PR #76, opened 2018-12-13
MODKBEKBJ-109: Sample of refactoring with Spring

andrii-paias:
## Purpose
Create example of how asynchronous call chain can be refactored to
only contain parts that are custom to each endpoint

## Approach
Extracted common logic into RMAPITemplate class,
Made conversion of some package classes automatic with ConversionService
class.


dmtkachenko:
I noticed that `CommonAttributesConverter` doesn't follow the same approach as other converters: convert from a single input data object to a single output.
Any reason for that?

PR #79, opened 2018-12-14
MODKBEKBJ-112 -  load and use module schemas in API tests for validat…

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-112 we want to have the ability to get JSON schemas for response validation used in postman API tests. 
this PR consists of two parts - adding system endpoints and refactoring tests, which will be linked.

## Approach
added `_jsonSchemas` and `_ramls` service endpoints

#### TODOS and Open Questions
- after this PR merged there is a second one related to api-tests which also should be reviewed. 

## Learning
* here is documentation describes system endpoints
 JSON Schema API - https://github.com/folio-org/raml-module-builder#json-schemas-api     
 RAML API - https://github.com/folio-org/raml-module-builder#ramls-api

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -126,6 +126,28 @@
           "permissionsRequired" :["kb-ebsco.cache.delete"]
         }
       ]
+    },
+    {
+      "id": "_jsonSchemas",
+      "version": "1.0",
+      "interfaceType" : "multiple",
+      "handlers" : [
+        {
+          "methods" : [ "GET" ],
+          "pathPattern" : "/_/jsonSchemas"
+        }
+      ]
+    },
+    {
+      "id": "_ramls",
+      "version": "1.0",
+      "interfaceType" : "multiple",
+      "handlers" : [
+        {
+          "methods" : [ "GET" ],
+          "pathPattern" : "/_/ramls"
+        }
+      ]
     }
   ],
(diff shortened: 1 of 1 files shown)


natali-zaitseva:
Hello @sduvvuri-ebsco. Those two endpoints `/_/jsonSchemas` and `/_/ramls` are part of the RMB module as well as the tenant endpoint  https://github.com/folio-org/okapi/blob/master/doc/guide.md#example-3-upgrading-versions-environment-and-the-_tenant-interface and they serve for the purpose, for instance, getting files associated with the module. I am sure that with our approach we did not violate the contract and for now, I do not really know why other modules did not use those endpoints, perhaps they faced with other challenges. We also can use a reference to GitHub to fetch the schema but I do not see the reason for it because developers/architectors are already taken care of this case and all we need is to use it correctly.  As for `$ref` - this is fixed and schemas, which are referenced that way are also downloadable.

natali-zaitseva:
@sduvvuri-ebsco Those endpoints were recently added(21 Nov 2018), probably this is the reason why they are not used in modules. I have created a Confluence page with small description https://wiki.folio.org/display/FOLIJET/RMB%3A+JSON+Schema+and+RAML+interfaces+overview, which may clarify setup and usage

sduvvuri-ebsco:
Thanks @natali-zaitseva !

PR #86, opened 2019-01-04
MODKBEKBJ-109  Simplify call to RM API configuration in composed processing chain

dmtkachenko:
## Purpose
JIRA: [MODKBEKBJ-109](https://issues.folio.org/browse/MODKBEKBJ-109)
Simplify call to RM API configuration in composed processing chain. 

## Approach
* Represent a call to RM API as a template where there are several steps:
  1. calling configuration service
  2. applying configuration settings to RM API Service
  3. execute RM API Service method (action in terms of the template)
  4. handle Success/Failure

3rd and 4th steps can be configured by template caller. 1st and 2nd are common and unchangeable

* Utilize [Spring Conversion service](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/convert/ConversionService.html) and [Converters](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/convert/converter/Converter.html) to build all the classes that transform REST object to RM API objects and otherwise around the common interface. This brings several benefits:
  - Converters are well structured (per business entity)
  - Top level converters include basic converters. These basic converters are independent and can be shared between several top level converters
  - All the converters registered in Conversion Service. Potentially it can be used as one and only thing required to transform from one object to another. In particular RM API template depends only on CS when it's required to build a response from RM API result object


src/main/java/org/folio/rest/impl/EholdingsProxyTypesImpl.java
```diff
@@ -1,97 +1,47 @@
...
-      });
+    rmapiTemplateFactory.createTemplate(okapiHeaders, asyncResultHandler)
+      .requestAction((rmapiService, okapiData) ->
+        rmapiService.retrieveProxies()
+      )
+      .addErrorMapper(RMAPIUnAuthorizedException.class, rmApiException -> GetEholdingsProxyTypesResponse
+        .status(HttpStatus.SC_FORBIDDEN)
+        .header(CONTENT_TYPE_HEADER, JSON_API_TYPE)
+        .entity(createError(rmApiException.getMessage()))
+        .build())
```
carolegodfrey: I think in general we could get an RMAPIUnAuthorizedException when calling any of the rm api related endpoints. It was explicitly added for this ProxyTypes endpoint but I'm wondering if the handling of this could be more generalized in the RMAPITemplate class?

src/main/java/org/folio/spring/config/ApplicationConfig.java
```diff
@@ -13,12 +18,20 @@
...
     configurer.setLocation(new ClassPathResource("application.properties"));
     return configurer;
   }
+
+  @Bean
+  public ConversionService conversionService(List<Converter> converters) {
+    DefaultConversionService conversionService = new DefaultConversionService();
+    converters.forEach(conversionService::addConverter);
+    return conversionService;
+  }
```
carolegodfrey: Would like to better understand why ConversionService is initialized here (with I assume all available converters injected)? As opposed to in TemplateFactory - Is this to guarantee a single instance of the class and converters?

src/main/java/org/folio/rest/impl/EholdingsProxyTypesImpl.java
dmtkachenko: This is probably more responsibility of `ErrorHandler`. It already contains some default logic:
image

If RMAPIUnAuthorizedException is common to all methods and should be handled in the same way, default mapper seems like a right place to do it

dmtkachenko: thanks @carolegodfrey 
I'll generalize the mapping

carolegodfrey: Thanks @dmtkachenko

dmtkachenko: refactored

PR #87, opened 2019-01-15
MODKBEKBJ-43: Use new endpoint to get VendorId for custom packages

andrii-paias:
## Purpose
Use updated "GET /rm/rmaccounts/{custid}" endpoint for getting vendor id of custom packages

## Approach
Change implementation to use GET /rm/rmaccounts/{custid} endpoint

#### TODOS and Open Questions
- [ ] Production environment (https://api.ebsco.io) returns vendor id correctly, but https://sandbox.ebsco.io currently returns following error 
```
{
    "errors": [
        {
            "code": "1004",
            "message": "IProxyClient#getCustomerProxies(String) timed-out and no fallback available.",
            "subCode": "0"
        }
    ]
}
```
So applying these changes will temporarily break get and update endpoints for custom packages on test environments.

### DIFF
--- a/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
+++ b/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
@@ -81,10 +81,15 @@ public class EholdingsPackagesImpl implements EholdingsPackages {
 
     templateFactory.createTemplate(okapiHeaders, asyncResultHandler)
       .requestAction((rmapiService, okapiData) ->
-        rmapiService.getVendors(isFilterCustom)
-          .thenCompose(vendors ->
-            rmapiService.retrievePackages(filterSelected, filterType, rmapiService.getFirstProviderElement(vendors),
-              q, page, count, nameSort)))
+      {
+        if (isFilterCustom) {
+          return rmapiService.getVendorId()
+            .thenCompose(vendorId ->
+              rmapiService.retrievePackages(filterSelected, filterType, vendorId, q, page, count, nameSort));
+        } else {
+          return rmapiService.retrievePackages(filterSelected, filterType, null, q, page, count, nameSort);
+        }
+      })
       .addErrorMapper(RMAPIServiceException.class,
         exception ->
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/rmapi/model/RootProxyCustomLabels.java
```diff
@@ -14,8 +14,11 @@
...
   private Proxy proxy;
-  
+
+  @JsonProperty("vendorId")
+  private String vendorId;
+
   @JsonProperty("labels")
   private List<CustomLabel> labelList;
-  
+
```
carolegodfrey: Was wondering on the name of this model -- it is now proxy, custom labels and vendor id (and will also have ftf tokens). It seems like it is a customer configuration.

carolegodfrey: Yes -- I agree - https://developer.ebsco.com/reference/rmapi#customlabelsproxy

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -81,10 +81,15 @@ public void getEholdingsPackages(String filterCustom, String q, String filterSel
 
     templateFactory.createTemplate(okapiHeaders, asyncResultHandler)
       .requestAction((rmapiService, okapiData) ->
-        rmapiService.getVendors(isFilterCustom)
-          .thenCompose(vendors ->
-            rmapiService.retrievePackages(filterSelected, filterType, rmapiService.getFirstProviderElement(vendors),
-              q, page, count, nameSort)))
+      {
+        if (isFilterCustom) {
+          return rmapiService.getVendorId()
```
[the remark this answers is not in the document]
carolegodfrey: Thanks for adding this caching @andrii-paias -- I think it is fine that VendorId is not invalidated as part of cache delete (this value should not be changing). Just curious on added RMAPITemplateContext - is this purpose primarily for simplifying usage in impl classes? (see a number of updates where we are now calling context.getService())

PR #88, opened 2019-01-16
MODKBEKBJ-146: PUT Root Proxy Updates to RM API Request

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-146 there have been updates to the following RM API endpoint rm/rmaccounts/{custid}

GET endpoint returns additional fields (VendorId and ftf tokens) and only returns custom labels which have a non empty display label value. Prior version returned an array of all custom labels (up to 5 including some with empty label values)

PUT request continues to support update of only proxy id and custom labels. Although there have been added fields in the GET response, they are not updated nor can be edited via the PUT. If included in the request payload they are simply ignored.

Mod-kb-ebsco-java's PUT root-proxy method had code which removed empty labels when making a request to update root-proxy. This is no longer necessary as RM API no longer returns empty values. This PR removes unnecessary code.
https://github.com/folio-org/mod-kb-ebsco-java/pull/88/commits/0b9175bd9a40c5d3cbf83a82af0086309b357575

## Approach
- Created as branch off of MODKBEKBJ-146-check-root-proxy
- Remove code which removed empty custom labels
- Remove no longer applicable unit test and associated test files

## Learning
Test of endpoint as documented at https://developer.ebsco.com/interactive/holdingsiq#/


carolegodfrey:
Created a separate branch off master instead

PR #91, opened 2019-01-22
MODKBEKBJ-126: Steps to replace mod-kb-ebsco Ruby with Java module

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-126 need to coordinate on the replacement process for mod-kb-ebsco (ruby module) with mod-kb-ebsco (Java module)

Work for doing the replacement is noted here:
https://issues.folio.org/browse/FOLIO-1727

NOTE: This PR should not be merged until we  are ready to swap out the ruby-based module in folio-snapshot environment. Initially swap will be done in folio/testing environment and tested before updating all environments.  (Q42018 environment is not impacted by this change and will continue to use mod-kb-ebsco (Ruby module)

## Approach
- Set Jenkinsfile publishModDescriptor = ‘yes’ 
- Upgrade version of mod-kb-ebsco-java to version 2.0.0-SNAPSHOT in pom file. It is should to be a later version than the current Ruby version of mod-kb-ebsco
- Update news.md to reflect the changes

## Learning

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-kb-ebsco-java</artifactId>
-  <version>0.1.0-SNAPSHOT</version>
+  <version>2.0.0-SNAPSHOT</version>
   <packaging>jar</packaging>
 
   <name>EBSCO KB Broker</name>
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 buildMvn {
-  publishModDescriptor = 'no'
+  publishModDescriptor = 'yes'
   mvnDeploy = 'yes'
   publishAPI = 'yes'
   runLintRamlCop = 'yes'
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -3,7 +3,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.folio</groupId>
   <artifactId>mod-kb-ebsco-java</artifactId>
-  <version>0.1.0-SNAPSHOT</version>
+  <version>2.0.0-SNAPSHOT</version>
```
natali-zaitseva: just curious about major version change from 0.1.0 to 2.0.0. probably my mean 1.0.0?

carolegodfrey: No -- good question @natali-zaitseva -- it is 2.0.0-SNAPSHOT as per suggestion/conversation with @funkymalc. The current plan is to make the version greater than the latest release version of mod-kb-ebsco. Some ongoing discussion around this here - https://issues.folio.org/browse/FOLIO-1727

natali-zaitseva: Thank you for explanation @carolegodfrey!

PR #94, opened 2019-01-29
FOLIO-1753 - Release Java module so it can be included environments

carolegodfrey:
## Purpose
Per https://issues.folio.org/browse/FOLIO-1753 we are releasing mod-kb-ebsco-java module so that it can be included in environments and we can start deprecating the use of mod-kb-ebsco (Ruby Version). Of note - the Ruby version will continue to be included as part of Q3 and Q4 releases. Of note -  The Java module provides eholdings interface version 1.0, the Ruby module provides eholdings interface version 0.1

## Approach
- Update version, news and set publishModDescriptor = yes
- Used `mvn -DautoVersionSubmodules=true release:clean release:prepare` so to properly handle release and snapshot versioning

#### TODOS and Open Questions
- [ ] A follow on release of eholdings is required. It will be updated to reference the newer 1.0 eholdings interface version

## Learning
Followed release procedures as noted at https://dev.folio.org/guidelines/release-procedures/#maven-based-modules

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 buildMvn {
-  publishModDescriptor = 'no'
+  publishModDescriptor = 'yes'
   mvnDeploy = 'yes'
   publishAPI = 'yes'
   runLintRamlCop = 'yes'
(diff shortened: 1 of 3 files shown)


carolegodfrey:
This was done in separate PR/branch

PR #96, opened 2019-02-01
MODKBEKBJ-167 - Create database script to store tag data

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-167 we want to be able to store tag data 

## Approach
* added /_/tenant endpoint to ModuleDescriptor file because database schema creation triggers on the step of enabling the module to a tenant.
* created a schema to store tags
* added tested data

## Learn
* Tenant Api - https://github.com/folio-org/raml-module-builder#tenant-api
screen shot 2019-02-01 at 11 32 37 am

* To enable test data creation we have to specify an environment variable called `test.mode` in DeploymentDescriptor.
screen shot 2019-02-04 at 11 35 08 am
the corresponding log file with message about test data creation can be viewed by 
`docker logs <docker_container_name>`
screen shot 2019-02-04 at 11 36 11 am

### DIFF
--- a/src/main/resources/templates/db_scripts/create-table.sql
+++ b/src/main/resources/templates/db_scripts/create-table.sql
@@ -1,0 +1,10 @@
+ALTER TABLE tags
+DROP COLUMN jsonb,
+ADD COLUMN record_id VARCHAR (50) NOT NULL,
+ADD COLUMN record_type VARCHAR (50) NOT NULL,
+ADD COLUMN tag VARCHAR (50) NOT NULL;
+
+INSERT INTO tags (record_id, record_type, tag) VALUES
+('583-4345-762169', 'resource', 'first tag'),
+('4345', 'package', 'second tag'),
+('762169', 'title', 'third tag');
--- a/src/main/resources/templates/db_scripts/create-table.sql
+++ b/src/main/resources/templates/db_scripts/create-table.sql
@@ -4,7 +4,5 @@ ADD COLUMN record_id VARCHAR (50) NOT NULL,
 ADD COLUMN record_type VARCHAR (50) NOT NULL,
 ADD COLUMN tag VARCHAR (50) NOT NULL;
 
-INSERT INTO tags (record_id, record_type, tag) VALUES
-('583-4345-762169', 'resource', 'first tag'),
-('4345', 'package', 'second tag'),
-('762169', 'title', 'third tag');
+CREATE INDEX record_id_index ON tags (record_id);
+CREATE INDEX tag_index ON tags (tag);
--- a/src/main/resources/templates/db_scripts/create-table.sql
+++ b/src/main/resources/templates/db_scripts/create-table.sql
@@ -1,7 +1,7 @@
 ALTER TABLE tags
 DROP COLUMN jsonb,
 ADD COLUMN record_id VARCHAR (50) NOT NULL,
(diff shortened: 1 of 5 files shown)


src/main/resources/templates/db_scripts/create-table.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE tags
+DROP COLUMN jsonb,
+ADD COLUMN record_id VARCHAR (50) NOT NULL,
+ADD COLUMN record_type VARCHAR (50) NOT NULL,
+ADD COLUMN tag VARCHAR (50) NOT NULL;
+
+INSERT INTO tags (record_id, record_type, tag) VALUES
+('583-4345-762169', 'resource', 'first tag'),
+('4345', 'package', 'second tag'),
+('762169', 'title', 'third tag');
```
carolegodfrey: Looks good Natalia
Just a few questions for understanding:
- assuming tables will be tenant specific - correct assumption?
- will we have a predefined set of record types (provider, package, resource and tile) for example?
- trying to understand if our package (record_id) would be of the format vendorId-packageId or package_id? -- I think vendorId-packageId will give us a unique package

natali-zaitseva: > assuming tables will be tenant specific - correct assumption?

the name of the table will be 'tags' but the name of the schema will consist of the following parts: `<tenant>_<moduleName>`
as an example, please see the following image
screen shot 2019-02-04 at 11 21 42 am 1

> will we have a predefined set of record types (provider, package, resource and tile) for example?

yes, we will have test data . The code for inserting data will be stored in separate file
`INSERT INTO tags (record_id, record_type, tag) VALUES
('583-4345-762169', 'resource', 'first tag'),
('4345', 'package', 'second tag'),
('762169', 'title', 'third tag');`
I will add possibility to add test data by environment flag in nex commit

> trying to understand if our package (record_id) would be of the format vendorId-packageId or package_id? -- I think vendorId-packageId will give us a unique package

here we can decide what is more convinient for us in case of search

src/main/resources/templates/db_scripts/test-data.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO ${myuniversity}_${mymodule}.tags (record_id, record_type, tag) VALUES
+('583-4345-762169', 'resource', 'EBSCO'),
```
dmtkachenko: please make sure that record ids are real

src/main/resources/templates/db_scripts/create-table.sql
dmtkachenko: It will be the same value that comes from UI: vendorId-packageId

carolegodfrey: Thanks for all of these details and explanation!

src/main/resources/templates/db_scripts/create-table.sql
```diff
@@ -1,7 +1,7 @@
 ALTER TABLE tags
 DROP COLUMN jsonb,
 ADD COLUMN record_id VARCHAR (50) NOT NULL,
-ADD COLUMN record_type VARCHAR (50) NOT NULL,
+ADD COLUMN record_type VARCHAR (10) CHECK (record_type IN ('provider', 'package', 'title', 'resource')),
```
dmtkachenko: :thumbsup:

PR #99, opened 2019-02-04
MODKBEKBJ-158 - Tags: retrieve tags assigned to Provider record

natali-zaitseva:
## Purpose
As the parent task https://issues.folio.org/browse/MODKBEKBJ-158 for two related sub-task, we want to add changes to the master for maintaining tags functionality

## Learning
* https://issues.folio.org/browse/MODKBEKBJ-167
* https://issues.folio.org/browse/MODKBEKBJ-166


src/main/java/org/folio/rest/impl/TenantApiImpl.java
```diff
@@ -0,0 +1,80 @@
...
+      } else {
+        setupTestData(headers, context).setHandler(event -> handlers.handle(result));
+      }
+    }, context);
+  }
+
+  private Future<List<String>> setupTestData(Map<String, String> headers, Context context) {
+    try {
+
+      if (!Boolean.TRUE.equals(Boolean.valueOf(System.getenv(TEST_MODE)))) {
```
carolegodfrey: Just curious -- is this a commonly used approach for controlling whether test data is loaded or not?

src/main/resources/templates/db_scripts/test-data.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO ${myuniversity}_${mymodule}.tags (record_id, record_type, tag) VALUES
+('53-2767121-90099', 'resource', 'EBSCO'),
+('53-2767121', 'package', 'folio'),
+('413-1988660', 'package', 'spitfire');
+('36-2728041', 'package', 'EBSCO');
```
carolegodfrey: :thumbsup: on package record_id(s)

ramls/types/providers/providerDataAttributes.json
```diff
@@ -28,13 +28,18 @@
...
     },
     "proxy": {
       "type": "object",
       "description": "Proxy",
       "$ref": "../proxy.json"
+    },
+    "tags": {
+      "type": "object",
+      "description": "Provider tags",
+      "$ref": "../../raml-util/schemas/tags.schema"
```
carolegodfrey: Just 1 clarification -- although tags are defined as part of attributes for provider here, we are populating values ONLY for get provider by id at this point (correct)? Hopefully this will not be any issue for UI as missing attributes will NOT be returned for list request.

natali-zaitseva: providerDataAttributes.json is used for retrieving provider by id and for Put provider

src/main/java/org/folio/rest/impl/TenantApiImpl.java
natali-zaitseva: it is a custom approach

ramls/types/providers/providerDataAttributes.json
carolegodfrey: Ok, thanks @natali-zaitseva -- just wanted to confirm that the added attribute does not cause issues for other methods which may be sharing the same definition (providerDataAttributes.json )

src/main/java/org/folio/rest/impl/TenantApiImpl.java
carolegodfrey: Ok, thanks -- so by default test data will not get loaded. Environment variable is needed for this as you noted in the sample deployment descriptor from previous PR -- got it

PR #105, opened 2019-02-06
MODKBEKBJ-161 - Tags: assign/unassign tags to Provider record

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-161 we want to have a functionality to assign and unassign tags for Provider entity

## Approach
 added assign/unassign functionality for Provider by PostgresClient

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -4,7 +4,7 @@
       "tableName": "tags",
       "withMetadata": false,
       "pkColumnName": "id",
-      "generateId": true,
+      "generateId": false,
       "withAuditing": false,
       "customSnippetPath": "create-table.sql"
     }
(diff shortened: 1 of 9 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -4,7 +4,7 @@
       "tableName": "tags",
       "withMetadata": false,
       "pkColumnName": "id",
-      "generateId": true,
+      "generateId": false,
```
carolegodfrey: Just curious as to why we changed this?

natali-zaitseva: I have found that we have test data duplication during the enabling/disabling module for the test purposes. As we leave PK creation to DB, the script will create new PKs for our entities and add them to already existing rows. Unfortunately, schema deletion is not performed on a disabling module for the tenant but performed on tenant deletion as it mentioned in doc https://github.com/folio-org/raml-module-builder#the-delete-tenant-api. This way is not convenient for us, that is why I have changed pk generation attribute to false.

carolegodfrey: Thanks for the detailed information @natali-zaitseva

PR #106, opened 2019-02-07
MODKBEKBJ-168: Fix bug where missing attributes caused error on PUT

andrii-paias:
Jira: MODKBEKBJ-168

## Purpose
Fix bug with PUT request for resource

## Approach
Merged existing resource attributes with attributes in PUT request.
If PUT request doesn't have an attribute then request to RM API will
be populated with old value for this attribute, so that the value will
not be changed

### DIFF
--- a/src/main/java/org/folio/rest/converter/providers/ProviderPutRequestConverter.java
+++ b/src/main/java/org/folio/rest/converter/providers/ProviderPutRequestConverter.java
@@ -1,14 +1,13 @@
 package org.folio.rest.converter.providers;
 
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.core.convert.converter.Converter;
-import org.springframework.lang.NonNull;
-import org.springframework.stereotype.Component;
-
 import org.folio.rest.jaxrs.model.PackageCollection;
 import org.folio.rest.jaxrs.model.ProviderPutRequest;
 import org.folio.rmapi.model.VendorPut;
 import org.folio.rmapi.model.VendorPutToken;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.lang.NonNull;
+import org.springframework.stereotype.Component;
 
 @Component
 public class ProviderPutRequestConverter implements Converter<ProviderPutRequest, VendorPut> {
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/rest/converter/providers/ProviderPutRequestConverter.java
```diff
@@ -24,7 +23,8 @@ public VendorPut convert(@NonNull ProviderPutRequest provider) {
     // value
     // Hard code it to false; it should not affect the state of inherited that RM
     // API maintains
-    if (provider.getData().getAttributes().getProxy() != null) {
+    if (provider.getData().getAttributes().getProxy() != null &&
```
andrii-paias: When updating provider, Ruby module actually retrieves old provider data and uses it if proxy or token are not set.
I think it is not necessary, if proxy or token are set, then UI will send them, and if it is not set, then we will just send null to RM API.

src/main/java/org/folio/rest/validator/ResourcePutBodyValidator.java
```diff
@@ -67,7 +68,10 @@ private void validateManagedResourceIfNotSelected(ResourceDataAttributes attribu
     Boolean isHidden = attributes.getVisibilityData() != null ? attributes.getVisibilityData().getIsHidden() : null;
     EmbargoUnit embargoUnit = attributes.getCustomEmbargoPeriod() != null ? attributes.getCustomEmbargoPeriod().getEmbargoUnit() : null;
     List<Coverage> customCoverages = attributes.getCustomCoverages();
-    if (!isTitleCustom && (!Objects.isNull(cvgStmt) || !Objects.isNull(embargoUnit) || (!Objects.isNull(isHidden) && isHidden) || (!Objects.isNull(customCoverages) && !customCoverages.isEmpty()))) {
+    if (!isTitleCustom &&
+      (!Strings.isEmpty(cvgStmt) || !Objects.isNull(embargoUnit) ||
+      (!Objects.isNull(isHidden) && isHidden) ||
+      (!Objects.isNull(customCoverages) && !customCoverages.isEmpty()))) {
       throw new InputValidationException(INVALID_IS_SELECTED_TITLE, INVALID_IS_SELECTED_DETAILS);
```
carolegodfrey: This LGTM @andrii-paias -- my main concern is the case where we deselect a resource -- we get errors from RM API if some of these fields are populated . This case should be caught here by validator

andrii-paias: yes, this should be caught by this condition, 
Here I only changed "!Objects.isNull(isHidden)" to "!Strings.isEmpty(cvgStmt)", RM API doesn't return an error if cvgStmt is an empty string, so we should let that case pass.

PR #108, opened 2019-02-11
MODKBEKBJ-163 assign tags to title

andrii-paias:
## Purpose
Allow assigning/unassigning tags to title with PUT titles/{id} endpoint

## Approach
Inserted tags into database after successful update of title on put.
Added small fixes to PUT titles/{id} endpoint


dmtkachenko:
It's important to have a common attribute for tags and not introduce object or even method specific naming. 
Title and resource modification has to be separated: titles have to have their own method to update data. Then it'd be possible to use a single tag. This will be addressed in MODKBEKBJ-177

PR #110, opened 2019-02-12
MODKBEKBJ-163 Assign title tags on post

andrii-paias:
## Purpose
Assign tags on POST /titles request

## Approach
Add tags to POST /titles schema, 
call method in TagRepository to save tags


src/main/java/org/folio/rmapi/model/Identifier.java
```diff
@@ -14,10 +14,6 @@
   @JsonProperty("id")
   public String id;
 
-  @JsonInclude(JsonInclude.Include.NON_NULL)
-  @JsonProperty("source")
-  public String source;
-
```
carolegodfrey: Just curious why this was here and now removed

andrii-paias: Identifier had source field just to map all fields that are returned from RM API, 
we sometimes get Identifiers from kb-ebsco-java model (which doesn't have source), and sometimes from RM API(which has source), 
to make it more consistent I removed this field

PR #113, opened 2019-02-26
MODCXEKB-90- Replace code for 'mod-kb-ebsco-java' with library

natali-zaitseva:
## Purpose
Due to the creation of the  library that contains all needed code to communicate with RMAPI, we should replace duplicated code with the library usage

## Approach
* replaced RMAPI code with library
* removed duplicated classes
* removed tests related to RMAPI code


pom.xml
```diff
@@ -166,6 +166,11 @@
       <artifactId>folio-di-support</artifactId>
       <version>1.0.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.folio</groupId>
+      <artifactId>folio-holdingsiq-client</artifactId>
+      <version>0.0.1-SNAPSHOT</version>
```
dmtkachenko: please update to released version when available

natali-zaitseva: will do right after the release

PR #131, opened 2019-04-19
MODKBEKBJ-209 - Do not capitalize contributor type

natali-zaitseva:
## Purpose
Updates for https://issues.folio.org/browse/MODKBEKBJ-209

## Approach
* updated ContributorsConverterPair to use a lowercase method
* updated tests

### DIFF
--- a/src/main/java/org/folio/rest/converter/common/attr/ContributorsConverterPair.java
+++ b/src/main/java/org/folio/rest/converter/common/attr/ContributorsConverterPair.java
@@ -6,7 +6,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 
-import org.apache.commons.lang.StringUtils;
 import org.springframework.core.convert.converter.Converter;
 import org.springframework.lang.NonNull;
 import org.springframework.lang.Nullable;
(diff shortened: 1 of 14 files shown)


natali-zaitseva:
@sduvvuri-ebsco as for `contributor type` - we use it, for instance, in POST /title and without this parameter RM API returns following error 
`{
    "errors": [
        {
            "title": "Parameter contributorsList.type can not be null or empty."
        }
    ],
    "jsonapi": {
        "version": "1.0"
    }
}`
I am not completely sure why UI has problems with a capitalized value of contributor type

sduvvuri-ebsco:
@natali-zaitseva : Probably, my earlier comment wasn't clear. I meant whether we are getting Contributor Type in upper case or lower case from RM API so that we can just pass it through to UI. Is the casing consistent? UI just wanted to avoid additional computation on the backend which is why this story was put together. I can explain better in a call right after our stand up if thats easier.

natali-zaitseva:
@sduvvuri-ebsco got it, thanks for the explanation. You are right, RM API returns values already in lower case, so additional call `lowerCase` for the type is not needed

sduvvuri-ebsco:
@natali-zaitseva : Perfect. Thanks for checking.

PR #132, opened 2019-04-23
MODKBEKBJ-210 Add packages table

andrii-paias:
## Purpose
Add table for storing package data locally

## Approach
Add table with name and contentType

#### TODOS and Open Questions
- [ ] Alternatively, we can create another table "record", which will act as a base table for tables packages, titles etc.
image
With this structure we will be able to use foreign keys and preserve referential integrity.
But inserting and updating record data will be more awkward, because we will need to insert data into multiple tables


andrii-paias:
@sduvvuri-ebsco, yes

PR #133, opened 2019-04-24
MODKBEKBJ-210 Save package when corresponding tags are updated

andrii-paias:
## Purpose
Populate packages table 

## Approach
Execute insert or update on packages table when package tags are updated
Add unit tests

#### TODOS and Open Questions
- [x] I'm not sure if we also need to delete package data from database if it doesn't doesn't have tags anymore


src/main/java/org/folio/tag/repository/TagRepositoryImpl.java
```diff
@@ -189,7 +190,7 @@ private String getTableName(String tenantId) {
   private CompletableFuture<Boolean> unAssignTags(AsyncResult<SQLConnection> connection, String tenantId, String recordId,
       RecordType recordType) {
     CompletableFuture<Boolean> future = new CompletableFuture<>();
-    final String deleteQuery = String.format(DELETE_TAG_RECORD, getTableName(tenantId), recordId, recordType);
```
andrii-paias: small fix here, 
DELETE_TAG_RECORD only has one "%s" placeholder,
recordId, recordType parameters are passed to postgresClient later

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -260,7 +264,11 @@ public void getEholdingsPackagesResourcesByPackageId(String packageId, String so
     if (tags == null){
       return CompletableFuture.completedFuture(new PackageResult(packageId, null, null));
     }else {
-      return tagRepository.updateRecordTags(tenant, packageId.getVendorId() + "-" + packageId.getPackageId(), RecordType.PACKAGE, tags.getTagList())
+      return
+        packageRepository.savePackage(packageId, tenant)
```
sduvvuri-ebsco: Can we rename packageId to packageData or packageByIdData instead because that is really what it is?

src/main/java/org/folio/tag/repository/packages/PackageRepositoryImpl.java
```diff
@@ -0,0 +1,73 @@
...
+      .thenApply(result -> null);
+  }
+
+  private JsonArray createInsertOrUpdateParameters(String id, String name, String contentType) {
+    JsonArray parameters = new JsonArray();
+    parameters
+      .add(id)
+      .add(name)
+      .add(contentType)
+      .add(name)
```
sduvvuri-ebsco: duplicate name and contentType?

src/test/resources/requests/rmapi/packages/post-package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "contentType": 1,
```
sduvvuri-ebsco: can we have a string content type please?

src/main/java/org/folio/tag/repository/packages/PackageRepositoryImpl.java
andrii-paias: Syntax of 
```INSERT_OR_UPDATE_STATEMENT =
    "INSERT INTO %s(" + PACKAGE_FIELD_LIST + ") VALUES (?, ?, ?) " +
      "ON CONFLICT (" + ID_COLUMN + ") DO UPDATE " +
      "SET " + NAME_COLUMN + " = ?, " + CONTENT_TYPE_COLUMN + " = ?";```
requires name and content type to be repeated

src/test/resources/requests/rmapi/packages/post-package.json
andrii-paias: That's a request ot RM API, so it has to have integer content type, 
This json was inlined in a test, so I moved it to a file

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
andrii-paias: Thanks, fixed

src/main/java/org/folio/tag/repository/packages/PackageRepositoryImpl.java
sduvvuri-ebsco: makes sense, my bad

src/test/resources/requests/rmapi/packages/post-package.json
sduvvuri-ebsco: makes sense, thanks for clarifying

PR #134, opened 2019-04-25
MODKBEKBJ-211 Add package cache

andrii-paias:
MODKBEKBJ-211
Add option to cache packages to PackageServiceImpl

## Purpose
Add caching for packages

## Approach
Add option to cache packages to PackageServiceImpl
Use VertxCache class to implement caching

### DIFF
--- a/src/main/java/org/folio/rmapi/PackageServiceImpl.java
+++ b/src/main/java/org/folio/rmapi/PackageServiceImpl.java
@@ -7,6 +7,7 @@ import java.util.concurrent.CompletableFuture;
 
 import io.vertx.core.Vertx;
 
+import org.folio.cache.VertxCache;
 import org.folio.holdingsiq.model.Configuration;
 import org.folio.holdingsiq.model.FilterQuery;
 import org.folio.holdingsiq.model.PackageByIdData;
@@ -15,6 +16,7 @@ import org.folio.holdingsiq.model.Sort;
 import org.folio.holdingsiq.model.Titles;
 import org.folio.holdingsiq.service.TitlesHoldingsIQService;
 import org.folio.holdingsiq.service.impl.PackagesHoldingsIQServiceImpl;
+import org.folio.rmapi.cache.PackageCacheKey;
 import org.folio.rmapi.result.PackageResult;
 import org.folio.rmapi.result.VendorResult;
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/rmapi/PackageServiceImpl.java
```diff
@@ -25,15 +27,31 @@
...
     this.providerService = providerService;
     this.titlesService = titlesService;
+    this.packageCache = packageCache;
+    this.tenantId = tenantId;
+    this.configuration = config;
   }
 
   public CompletableFuture<PackageResult> retrievePackage(PackageId packageId, List<String> includedObjects) {
-    CompletableFuture<PackageByIdData> packageFuture = retrievePackage(packageId);
+    return retrievePackage(packageId, includedObjects, false);
```
sduvvuri-ebsco: I am just wondering can we make the default lookup of a package in cache? If its not present in cache, we anyways reach out to RM API. What do you think?

src/main/java/org/folio/rmapi/PackageServiceImpl.java
```diff
@@ -25,15 +27,31 @@
...
+  }
+
+  public CompletableFuture<PackageResult> retrievePackage(PackageId packageId, List<String> includedObjects, boolean useCache) {
+    CompletableFuture<PackageByIdData> packageFuture;
+    if(useCache){
+      packageFuture = retrievePackageWithCache(packageId);
```
sduvvuri-ebsco: can we rename this to retrievePackageFromCache instead? Thanks!

src/main/java/org/folio/rmapi/PackageServiceImpl.java
andrii-paias: We can use cache by default, but I wanted to use cache only in case when we are getting tagged packages, 
then all other features will still have live data, and cache will be smaller, because it will only contain tagged packages

src/main/java/org/folio/rmapi/PackageServiceImpl.java
andrii-paias: I'm not sure, to me retrievePackageFromCache would mean that it returns null on cache miss, while this method loads package from RM API on cache miss

src/main/java/org/folio/rmapi/PackageServiceImpl.java
sduvvuri-ebsco: ok, makes sense 👍

src/main/java/org/folio/rmapi/PackageServiceImpl.java
sduvvuri-ebsco: Hmm - ok - just a naming thing. guess this is ok.

PR #135, opened 2019-04-30
MODKBEKBJ-188 Implement search of packages by list of tags

andrii-paias:
## Purpose
Implement search of packages by tags

## Approach
Add filter[tags] to GET /packages endpoint
Get list of package ids by tags
Get count of tagged packages
Send multiple asynchronous requests to RM API to get tagged packages


src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -226,6 +239,24 @@ public void getEholdingsPackagesResourcesByPackageId(String packageId, String so
...
+    MutableObject<Integer> totalResults = new MutableObject<>();
+    String tenant = context.getOkapiData().getTenant();
+    return tagRepository
+      .countRecordsByTags(tags, tenant, RecordType.PACKAGE)
+      .thenCompose(packageCount -> {
+        totalResults.setValue(packageCount);
+        return packageRepository.getPackageIdsByTagName(tags, page, count, tenant);
+      })
+      .thenCompose(packageIds ->
+        context.getPackagesService().retrievePackages(packageIds))
```
carolegodfrey: If 1 or more of the individual package requests fail -- will we fail the entire request? Probably an edge case (and maybe a separate issue/PR to consider)-- but if a package is no longer available in RM API -- want to make sure we will be able to handle

andrii-paias: Thanks, good point, 
for now added logging of errors returned from RM API, so if package isn't added then it will just not be returned in the response 
I'm not sure what we can do in this case, if it is possible for package to be removed from RM API, then we can handle the case of 404 error by deleting package and its tags from local database

sduvvuri-ebsco: I think its ok for the behavior to be - in a bunch of requests that we send to RM API - if we get a 404, then just don't display it. Deleting a package can happen only in case of custom packages that can result in a 404 from RM API. I am assuming the number should not be too many to increase the size of orphan tags in our tables. @carolegodfrey : Feel free to let us know your thoughts as well.

PR #136, opened 2019-05-02
MODKBEKBJ-215 Add cache for providers

andrii-paias:
## Purpose
Implement provider cache

## Approach
Implement vendor cache
Move construction of services to RMAPITemplateContextBuilder
Declare RMAPITemplateContextBuilder with prototype scope, so that
it is created on every request


src/main/java/org/folio/rest/util/template/RMAPITemplateContext.java
```diff
@@ -12,7 +11,6 @@
 import org.folio.rmapi.ResourcesServiceImpl;
 
 @Value
-@Builder(toBuilder = true)
```
carolegodfrey: Just curious -- you removed the `toBuilder = true` annotation -- which looks like it would have created RMAPITemplateContextBuilder.java on your behalf. Has this been removed because you now need access to the RMAPITemplateContextBuilder directly?

andrii-paias: I needed to add some logic to builder, so standard lombok implementation wouldn't work

Previously RMAPITemplateFactory had to pass all dependencies (e.g. cache objects) into RMAPITemplate, and RMAPITemplate then would pass them into builder. 
I wanted to inject depedencies directly into builder, and move logic for creating services into it as well, so now RMAPITemplate only has to pass Configuration and OkapiData to builder.

carolegodfrey: Ok thanks for details @andrii-paias

PR #140, opened 2019-05-06
MODKBEKBJ-214 Tags: add provider table to database

stanislav-mureiko:
## Purpose
https://issues.folio.org/browse/MODKBEKBJ-214
Implementing storing providers information in database.

## Approach
Adding provider repository.
Changing db schema.
Transferring classes to right packages.
Refactoring tests.


src/test/java/org/folio/util/TagsTestUtil.java
```diff
@@ -1,7 +1,20 @@
...
+import org.folio.tag.RecordType;
+import org.folio.tag.Tag;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+
+import static org.apache.commons.lang3.StringUtils.join;
```
carolegodfrey: @stanislav-mureiko -- this file as well as others only have changes due to ordering of imports -- I don't believe this ordering should be changed. If using import ordering configuration files for Intellij (provided by @dmtkachenko) -- order will and should not change

stanislav-mureiko: Thanks, @carolegodfrey -- I believe I was using configuration files for Intellij provided by @dmtkachenko. But I need to double check it.

stanislav-mureiko: I've fixed it in another PR https://github.com/folio-org/mod-kb-ebsco-java/pull/141

carolegodfrey: Ok thanks @stanislav-mureiko

PR #146, opened 2019-05-14
MODKBEKBJ-220 - Create holdings table

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-220 we want to save dump holdings data into the corresponding table.

## Approach
* added table to store holdings

#### TODOS and Open Questions
- [ ] Do we need to add some indexes?


src/main/resources/templates/db_scripts/create-holdings-table.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE holdings
+ALTER COLUMN id TYPE VARCHAR (50);
```
carolegodfrey: Assume we will be adding more columns which define resource here (placeholder for create)?

carolegodfrey: Understand all fields to be stored in jsonb column

PR #148, opened 2019-05-16
MODKBEKBJ-236 get provider packages by tags

andrii-paias:
## Purpose
Implement search of packages within providers by list of tags

## Approach
Reuse existing queries, but add ability to filter by provider id prefix with "LIKE {providerId}%"
Move common test methods and test data to common classes


src/main/java/org/folio/tag/repository/packages/PackageRepositoryImpl.java
```diff
@@ -75,11 +76,22 @@ public PackageRepositoryImpl(Vertx vertx, IdParser idParser) {
...
+    return getPackageIdsByTagAndIdPrefix(tags, providerId, page, count, tenant);
+  }
+
+  private CompletableFuture<List<PackageId>> getPackageIdsByTagAndIdPrefix(List<String> tags, String providerId, int page, int count, String tenantId) {
     int offset = (page - 1) * count;
 
     JsonArray parameters = new JsonArray();
     tags.forEach(parameters::add);
+    String likeExpression = providerId + "%";
     parameters
```
carolegodfrey: When filtering by providerId -- should like be for `providerId-` -- would not want to get provider id 18 and 181 results for example when searching for provider id 18.

andrii-paias: Thanks

PR #149, opened 2019-05-17
MODKBEKBJ-240 load tags for packages within providers

andrii-paias:
## Purpose
Return tags in response on search for packages within providers

## Approach
Load list of packages with tags from database and use it to populate response

#### TODOS and Open Questions
- [x] This PR depends on following PR: 
https://github.com/folio-org/mod-kb-ebsco-java/pull/148


src/main/java/org/folio/rest/converter/packages/PackageCollectionResultConverter.java
```diff
@@ -0,0 +1,59 @@
...
+  public PackageCollection convert(@NonNull PackageCollectionResult packagesResult) {
+    Packages packages = packagesResult.getPackages();
+    List<DbPackage> dbPackages = packagesResult.getDbPackages();
+    List<PackageCollectionItem> packageList = mapItems(packages.getPackagesList(),
+      packageData -> {
+        PackageCollectionItem item = packageCollectionItemConverter.convert(packageData);
+        item.getAttributes()
+          .withTags(new Tags()
+            .withTagList(getTagsById(dbPackages, createPackageId(packageData))));
+        return item;
```
carolegodfrey: @andrii-paias -- will tags now get included for all package collections in general? Realize the JIra issue was specific for provider (search packages within) -- but will tags also be calculated/returned for a general package list as well

andrii-paias: @carolegodfrey, 
no, they will only be returned if endpoint returns PackageCollectionResult (class with tag information), so now only packages withing providers have tags returned

PR #150, opened 2019-05-24
MODKBEKBJ-238 - create periodic task to populate holdings table

natali-zaitseva:
## Purpose
Due to https://issues.folio.org/browse/MODKBEKBJ-238 we want to have the ability to load holdings records for future use of the module.

## Approach
* added "_timer" system interface to run a periodic task to ModuleDescriptor file and assigned appropriate permission to get RM API configuration.
* added the corresponding endpoint to raml file
* added implementation of the scheduled task
* changed the version of jackson-databind to 2.9.9 due to a vulnerability issue 

## Learning
* Timer interface - https://github.com/folio-org/okapi/blob/master/doc/guide.md#timer-interface 
* KBART documentation - https://groups.niso.org/apps/group_public/download.php/12719/rp-9-2014_KBART.pdf

Need to be merged after #https://github.com/folio-org/folio-holdingsiq-client/pull/19


descriptors/ModuleDescriptor-template.json
```diff
@@ -174,6 +179,20 @@
...
+      "version": "1.0",
+      "interfaceType": "system",
+      "handlers": [
+        {
+          "methods": [ "POST" ],
+          "pathPattern": "/loadHoldings",
+          "modulePermissions": ["configuration.entries.collection.get"],
+          "unit": "day",
+          "delay": "2"
+        }
```
carolegodfrey: Didn't realize timer functionality is available -- do we have any way to control when (for example time of day) this periodic task will happen

natali-zaitseva: @carolegodfrey we are not able to specify the exact time, the time of the run will count from the module enabled for the tenant. For instance, module enabled at 24 May 2019 09:21:54 then the first run of the scheduled task will be 26 May 2019 09:21:54

ramls/eholdings.raml
```diff
@@ -94,3 +94,14 @@ types:
...
+/loadHoldings:
+  displayName: Run load holdings job
+  post:
+    description: Run load holdings job.
+    headers:
+      Content-Type:
+        example: application/json
+    responses:
+      204:
+        description: No Content
```
carolegodfrey: Thinking of debugging/troubleshooting this task (since it has multiple steps with potentials for issues/errors along the way - wondering if a separate endpoint for diagnostic information may be helpful. Realize there is a holdingsIq endpoint to check status and our log files/database (but not always easily accessible)

descriptors/ModuleDescriptor-template.json
natali-zaitseva: as for the future improvement we can do following - after the first time of setting RM API configuration we can do a call to load holdings, no matter when the module was enabled.

src/main/java/org/folio/rest/service/holdings/HoldingsServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+  @Autowired
+  public HoldingsServiceImpl(Vertx vertx, HoldingsRepository holdingsRepository, @Value("${holdings.status.check.delay}") long delay) {
+    this.vertx = vertx;
+    this.holdingsRepository = holdingsRepository;
+    this.delay = delay;
+  }
+
+  public CompletableFuture<Void> loadHoldings(RMAPITemplateContext context, String tenantId) {
+    return populateHoldings(context)
+      .thenCompose(isSuccessful -> waitForCompleteStatus(context, 5, tenantId))
```
carolegodfrey: Should parameters such as (5) be store in application.properties (as you did for delay)

descriptors/ModuleDescriptor-template.json
carolegodfrey: Thanks for clarification @natali-zaitseva

carolegodfrey:
Looks really good @natali-zaitseva -- my main concern would be for failures along the way for this process, having visibility into them and having a way to recover (if needed)

As I understand, each time the process runs it will update or add new records to the table (wonder if there are cases when we would want it to start fresh). For example — if a record gets removed from the set of holdings results,  will we be able to detect that?

Was also looking at documentation at 
https://developer.ebsco.com/interactive/holdingsiq#/
for `POST /{custid}/holdings` there is a listed a `force` parameter -- I assume this should only be applied as a manual step

natali-zaitseva:
@carolegodfrey as for 
> For example — if a record gets removed from the set of holdings results, will we be able to detect that?
--> we will download all that is in staging server, so if some record was deleted before our loading process it is just will not be present in our holdings table

carolegodfrey:
> @carolegodfrey as for
> 
> > For example — if a record gets removed from the set of holdings results, will we be able to detect that?
> 
> --> we will download all that is in staging server, so if some record was deleted before our loading process it is just will not be present in our holdings table

Ok, thanks for clarification @natali-zaitseva -- so is the table cleared prior to loading a new set of results from holdings? It looked as if it were adding/updating records and that is why I asked on the detection of a deleted record

natali-zaitseva:
@carolegodfrey, I am about to add deletion from the table before loading

ramls/eholdings.raml
dmtkachenko: > 
> 
> Thinking of debugging/troubleshooting this task (since it has multiple steps with potentials for issues/errors along the way - wondering if a separate endpoint for diagnostic information may be helpful. Realize there is a holdingsIq endpoint to check status and our log files/database (but not always easily accessible)

Agree. It's not easy to understand the progress and final result of data loading. Definitely a point to improve in future

PR #151, opened 2019-05-27
MODKBEKBJ-238: Save holdings in batches

andrii-paias:
## Purpose
Optimize saving holdings into local database

## Approach
Insert list of holdings in one transaction, by sending INSERT commands
with 200 holdings in each command.
Move common transaction logic into DBUtil.executeInTransaction method


src/main/java/org/folio/rest/repository/holdings/HoldingsRepositoryImpl.java
```diff
@@ -21,11 +28,13 @@
...
-  public static final String HOLDINGS_FIELD_LIST = String.format("%s, %s", ID_COLUMN, JSONB_COLUMN);
-  public static final String INSERT_OR_UPDATE_HOLDINGS_STATEMENT = "INSERT INTO %s(" + HOLDINGS_FIELD_LIST
-      + ") VALUES (?, ?) " + "ON CONFLICT (" + ID_COLUMN + ") DO UPDATE " + "SET " + JSONB_COLUMN + " = ?;";
+  private static final String ID_COLUMN = "id";
+  private static final String JSONB_COLUMN = "jsonb";
+  private static final String HOLDINGS_FIELD_LIST = String.format("%s, %s", ID_COLUMN, JSONB_COLUMN);
+  private static final String INSERT_OR_UPDATE_HOLDINGS_STATEMENT =
+    "INSERT INTO %s(" + HOLDINGS_FIELD_LIST + ") VALUES %s" +
+      "ON CONFLICT (" + ID_COLUMN + ") DO NOTHING";
+  private static final int MAX_BATCH_SIZE = 200;
```
carolegodfrey: One concern/question with the overall process -- in our holdings table will we not be able to detect that a record has been deleted from holdings?

sduvvuri-ebsco: why was the query changed from update to do nothing on conflict?

andrii-paias: @sduvvuri-ebsco,
I changed this in my commit, 
/holdings can return duplicated resources, (for example first two titles that I currently get have the same title/package ids), "on conflict do update" only works if duplicated rows are added in different "INSERT..." commands, when I tried to add same holding multiple times in one insert I got error "on conflict do update command cannot affect row a second time"

sduvvuri-ebsco: got it, thanks

PR #153, opened 2019-05-28
MODKBEKBJ-230: Filter titles by resource tags

andrii-paias:
## Purpose
Enable filtering titles by tags

## Approach
1)Calculate totalResults by counting tagged resources with distinct title part of id
2)Find titles that are associated with at least one resource that has one of
specified tags
3)For every found title load corresponding title data from RM API, if local database
doesn't already contain title data

When title is retrieved from local database then only following fields are returned:
 title id, title name, publication type, publisher names
Holding object has some other fields but it is tricky to map them correctly to title response

Move common test set up data into common classes


src/main/java/org/folio/repository/titles/TitlesRepository.java
```diff
@@ -1,12 +1,18 @@
...
 import org.folio.holdingsiq.model.Title;
+import org.folio.tag.repository.titles.DbTitle;
 
 public interface TitlesRepository {
 
   CompletableFuture<Void> saveTitle(Title title, String tenantId);
 
   CompletableFuture<Void> deleteTitle(String titleId, String tenantId);
+
+  CompletableFuture<List<DbTitle>> getTitlesByResourceTags(List<String> tags, int page, int count, String tenant);
```
sduvvuri-ebsco: Can the tenant be renamed to tenantId similar to other method params in this class for consistency? Thanks!

src/main/java/org/folio/repository/titles/TitlesRepository.java
```diff
@@ -1,12 +1,18 @@
...
+
+  CompletableFuture<Integer> countTitlesByResourceTags(List<String> tags, String tenant);
```
sduvvuri-ebsco: Same as above

src/main/java/org/folio/repository/titles/TitlesRepositoryImpl.java
```diff
@@ -61,7 +77,102 @@ public TitlesRepositoryImpl(Vertx vertx) {
...
+    }
+  }
+
+  private Title mapHoldingToTitle(Holding holding) {
+    return Title.builder()
+      .titleName(holding.getPublicationTitle())
+      .titleId(Integer.parseInt(holding.getTitleId()))
+      .pubType(holding.getResourceType())
+      .publisherName(holding.getPublisherName())
+      .build();
```
sduvvuri-ebsco: Since we are really using only a handful of entries from the holdings table, does it make sense to pre-process those entries and load only these into the DB table to reduce insertion time into DB or will pre-processing take more time than actual insertion into DB?

andrii-paias: Yes we can only store fields that we need from /holdings, 
will update this PR with these changes

sduvvuri-ebsco: ty @andrii-paias

PR #155, opened 2019-05-31
MODKBEKBJ-242 Title | List of Packages | Display assigned resource tags

stanislav-mureiko:
## Purpose
Displaying included into title resources with tags.
https://issues.folio.org/browse/MODKBEKBJ-242

## Approach
Adding new request to DB. Attaching tags from DB to included into title resource.
Adding and adjusting tests.

### DIFF
--- a/src/main/java/org/folio/repository/tag/TagRepositoryImpl.java
+++ b/src/main/java/org/folio/repository/tag/TagRepositoryImpl.java
@@ -105,8 +105,7 @@ class TagRepositoryImpl implements TagRepository {
   public CompletableFuture<List<Tag>> findByRecordByIds(String tenantId, List<String> recordIds, RecordType recordType) {
 
     String placeholders = createPlaceholders(recordIds.size());
-    recordIds.add(recordType.getValue());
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/repository/tag/TagRepositoryImpl.java
```diff
@@ -100,6 +101,22 @@ public TagRepositoryImpl(Vertx vertx) {
     return mapResult(resultSetFuture, this::readTags);
   }
 
+  @Override
+  public CompletableFuture<List<Tag>> findByRecordByIds(String tenantId, List<String> recordIds, RecordType recordType) {
+
+    String placeholders = createPlaceholders(recordIds.size());
+    recordIds.add(recordType.getValue());
```
dmtkachenko: Please avoid incoming list modification. Treat it is read only

stanislav-mureiko: Thanks, @dmtkachenko . I will fix it.

PR #156, opened 2019-05-31
MODKBEKBJ-237 - Tags: Update packages (search within titles ) to filt…

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-237 update existing implementation by adding the possibility to filter resources tied to package by tags

## Approach
* updated packages raml file for GET /eholdings/packages/{packageId}/resources
* updated implementation of the corresponding method

### DIFF
--- a/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
+++ b/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
@@ -1,12 +1,13 @@
 package org.folio.rest.impl;
 
+import static org.folio.common.ListUtils.mapItems;
+
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
-import java.util.stream.Collectors;
 
 import javax.validation.ValidationException;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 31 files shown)


src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -218,29 +225,73 @@ public void deleteEholdingsPackagesByPackageId(String packageId, Map<String, Str
...
+  private CompletableFuture<ResourceCollectionResult> getTitlesByPackageIdAndTags(String packageId, List<String> tags, int page, int count, RMAPITemplateContext context) {
+    MutableObject<Integer> totalResults = new MutableObject<>();
+    MutableObject<List<DbResource>> mutableDbTitles = new MutableObject<>();
+    String tenant = context.getOkapiData().getTenant();
+
+    return tagRepository
+      .countRecordsByTagsAndPrefix(tags, packageId + "-", tenant, RecordType.RESOURCE)
+      .thenCompose(resourceCount -> {
+        totalResults.setValue(resourceCount);
+        return resourceRepository.getResourcesByTagNameAndPackageId(tags, packageId, page, count, tenant);
```
andrii-paias: We can load holdings data for each resource from local db,
then we can only send requests to RM API for those resources that are missing from local db.

natali-zaitseva: @andrii-paias thanks for the note, added

PR #160, opened 2019-06-19
MODKBEKBJ-251 Provider: Make unit tests payloads more realistic

stanislav-mureiko:
## Purpose
Making tests more realistic

## Approach
Fixing tests records. Fixing several fields in test data

## Learning
I have looked through tests data and found out that there is no big difference between tests and real requests. Screenshots and comments you can find in Jira ticket https://issues.folio.org/browse/MODKBEKBJ-251


stanislav-mureiko:
@sduvvuri-ebsco  Actually we are using record with id 19 in tests. Through PUT we send id by url. For example eholdings/providers/19. And "id" is not necessary in body in this case. For changing other fields --
I have found realistic record from UI and changed fields in test record the same. You can see comment regarding this in Jira ticket https://issues.folio.org/browse/MODKBEKBJ-251

sduvvuri-ebsco:
ok, thanks for the explanation @stanislav-mureiko

PR #161, opened 2019-06-20
MODKBEKBJ-273 - Holdings: Update save holdings flow

natali-zaitseva:
## Purpose
Due to https://issues.folio.org/browse/MODKBEKBJ-273 we want to take steps to harden the process of loading holdings 

## Approach
* updated holdings table to include an additional column - decided not to use `DEFAULT CURRENT_TIMESTAMP` value due to the number of transactions performed. In this case, each transaction will have it is own value, which complicates the calculation of outdated entries. Instead, we set up a single value to all entries updated during the loading process. 
* updated SQL query used for an upsert statement - modified query to update holding value and time instead of ignoring.
* refactored repositories method naming
* updated tests

It was found during testing that RM API can return duplicates of the holdings. At the picture 
holdings size returned from RM API = 4831
Screen Shot 2019-06-19 at 5 29 44 PM
To avoid duplications the structure to carry holdings was changed to Set. The size of unique holdings at the picture  = 4829

## TODO
* To minimize interactions with the database we can think about reducing the time of loading from 2 days to 5-7 days or more.


src/main/java/org/folio/common/ListUtils.java
```diff
@@ -21,7 +21,7 @@ public static String createPlaceholders(int size) {
     return String.join(",", Collections.nCopies(size, "?"));
   }
 
-  public static  String createInsertPlaceholders(int size) {
-    return String.join(",", Collections.nCopies(size,"(?,?)"));
+  public static  String createInsertThreePlaceholders(int size) {
```
dmtkachenko: What if we need 4 placeholders or 5?

src/main/java/org/folio/repository/holdings/DbHolding.java
```diff
@@ -12,6 +12,7 @@
 )
 @Getter
 @Setter
+@EqualsAndHashCode
```
dmtkachenko:
```
@ToString
```
it's always nice to have toString() method also

src/main/java/org/folio/service/holdings/HoldingsServiceImpl.java
```diff
@@ -103,13 +103,15 @@ public void waitForCompleteStatus(RMAPITemplateContext context, int retries, Com
...
+    final Instant updatedAt = Instant.now();
     final int totalRequestCount = getRequestCount(totalCount);
     for (int iteration = 1; iteration < totalRequestCount + 1; iteration++) {
       int finalIteration = iteration;
       future = future
         .thenCompose(o -> context.getLoadingService().loadHoldings(MAX_COUNT, finalIteration))
-        .thenCompose(holding -> saveHolding(holding.getHoldingsList(), tenantId));
+        .thenCompose(holding -> saveHoldings(holding.getHoldingsList(), updatedAt, tenantId));
     }
+    future = future.thenCompose(o -> holdingsRepository.deleteByTimeStamp(updatedAt, tenantId));
```
dmtkachenko: Could you please explain why this is required?

natali-zaitseva: sure, @dmtkachenko let's assume we performed first loading and the time stamp of the operation is `2019-06-19T09:28:33.127+00`, so the next scheduled date to load holdings is + 2 days -> `2019-06-21T09:28:33.127+00`. The holdings entries which were removed from RM API between those two days will remain in our table with the time stamp of the previous iteration. This operation `holdingsRepository.deleteByTimeStamp()` removes outdated holding entries.

src/main/java/org/folio/repository/holdings/DbHolding.java
natali-zaitseva: thanks, will add

src/main/java/org/folio/service/holdings/HoldingsServiceImpl.java
dmtkachenko: in you sample, updatedAt == 2019-06-21T09:28:33.127+00, correct?

`deleteByTimeStamp()` removes entries from the previous run (records those timestamp is not equal to updatedAt), right?

natali-zaitseva: > deleteByTimeStamp() removes entries from the previous run (records those timestamp is not equal to updatedAt), right?

yes

src/main/java/org/folio/common/ListUtils.java
natali-zaitseva: @dmtkachenko refactored method to be more flexible

src/main/java/org/folio/service/holdings/HoldingsServiceImpl.java
```diff
@@ -131,8 +133,8 @@ private int getRequestCount(Integer totalCount) {
     return context.getLoadingService().getLoadingStatus();
   }
 
-  private CompletableFuture<Void> saveHolding(List<Holding> holdings, String tenantId) {
-    List<DbHolding> dbHoldings = holdings.stream()
+  private CompletableFuture<Void> saveHoldings(List<Holding> holdings, Instant updatedAt, String tenantId) {
+    Set<DbHolding> dbHoldings = holdings.stream()
```
andrii-paias: I'm still worried that there will be holding objects with same id, but different other attributes,
I think it's better to remove duplicates by using only id.
something like this 
```
  private CompletableFuture<Void> saveHoldings(List<Holding> holdings, Instant updatedAt, String tenantId) {
    Set<DbHolding> dbHoldings = holdings.stream()
      .filter(distinctByKey(this::getHoldingsId))
      .map(holding -> new DbHolding(
        holding.getTitleId(),
        holding.getPackageId(),
        holding.getVendorId(),
        holding.getPublicationTitle(),
        holding.getPublisherName(),
        holding.getResourceType()
      ))
      .collect(Collectors.toSet());
    logger.info("Saving holdings to database.");
    return holdingsRepository.saveAll(dbHoldings, updatedAt, tenantId);
  }

  private  <T> Predicate<T> distinctByKey(Function<? super T, ?> keyExtractor) {
    Set<Object> seen = ConcurrentHashMap.newKeySet();
    return t -> seen.add(keyExtractor.apply(t));
  }

  private String getHoldingsId(Holding holding) {
    return holding.getVendorId() + "-" + holding.getPackageId() + "-" + holding.getTitleId();
  }
```

natali-zaitseva: thanks @andrii-paias, added

PR #164, opened 2019-06-26
MODKBEKBJ-280 Add separate endpoint for package tags

andrii-paias:
## Purpose
Provide a separate endpoint for updating tags

## Approach
Add schema for PUT "packages/{packageId}/tags" endpoint and implement it
Add unit tests


src/main/java/org/folio/repository/packages/PackageRepository.java
```diff
@@ -3,11 +3,10 @@
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 
-import org.folio.holdingsiq.model.PackageByIdData;
 import org.folio.holdingsiq.model.PackageId;
 
 public interface PackageRepository {
-  CompletableFuture<Void> save(PackageByIdData packageData, String tenantId);
+  CompletableFuture<Void> save(DbPackage packageData, String tenantId);
```
sduvvuri-ebsco: Can we please rename DbPackage to something else - probably packageByIdDataInDB?

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -178,7 +193,9 @@ public void postEholdingsPackages(String contentType, PackagePostRequest entity,
       .requestAction(context ->
         getVendorId(context)
           .thenCompose(id -> context.getPackagesService().postPackage(packagePost, id))
-          .thenCompose(packageById -> updateTags(packageById, context.getOkapiData().getTenant(), tags))
+          .thenCompose(packageById ->
```
sduvvuri-ebsco: @andrii-paias : Confused as to why we are dealing with tags on POST package endpoint at all? When I try to create a new package(which is the only time a POST is being invoked), the only fields that I see are name, content type and coverage settings. I don't think we have to do anything with tags at all in the POST request unless I m missing something.

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -397,41 +449,49 @@ private String getResourceId(TitleResult titleResult) {
...
-    packageRepository.delete(packageId, tenant)
-    .thenCompose(o -> tagRepository.deleteRecordTags(tenant, packageId.getProviderIdPart() + "-" + packageId.getPackageIdPart(), RecordType.PACKAGE))
-      .thenCompose(aBoolean ->  completedFuture(null));
+      packageRepository.delete(packageId, tenant)
+        .thenCompose(o -> tagRepository.deleteRecordTags(tenant, packageId.getProviderIdPart() + "-" + packageId.getPackageIdPart(), RecordType.PACKAGE))
+        .thenCompose(aBoolean -> completedFuture(null));
   }
 
-  private CompletableFuture<PackageResult> updateTags(PackageByIdData packageData, String tenant, Tags tags) {
+  private CompletableFuture<Void> updateTags(Tags tags, DbPackage dbPackage, String tenant) {
```
sduvvuri-ebsco: After we update tags, are we not supposed to return updated tag list or empty list in case there are no tags? Is it right to return <Void> here?

src/test/java/org/folio/rest/validator/PackageTagsPutBodyValidatorTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.folio.rest.jaxrs.model.ContentType;
+import org.folio.rest.jaxrs.model.PackageTagsDataAttributes;
+import org.folio.rest.jaxrs.model.PackageTagsPutData;
+import org.folio.rest.jaxrs.model.PackageTagsPutRequest;
+
+public class PackageTagsPutBodyValidatorTest {
+  private PackageTagsPutBodyValidator validator = new PackageTagsPutBodyValidator();
+
+  @Test(expected = InputValidationException.class)
+  public void shouldTrowExceptionWhenNameIsEmpty(){
```
sduvvuri-ebsco: typo on "throw"

src/test/java/org/folio/rest/validator/PackageTagsPutBodyValidatorTest.java
```diff
@@ -0,0 +1,35 @@
...
+      .withData(new PackageTagsPutData()
+        .withAttributes(new PackageTagsDataAttributes()
+          .withName("")
+          .withContentType(ContentType.E_BOOK)
+        ));
+    validator.validate(request);
+  }
+
+  @Test(expected = InputValidationException.class)
+  public void shouldTrowExceptionWhenContentTypeIsNull(){
```
sduvvuri-ebsco: typo on throw

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
andrii-paias: If update completed successfully then the new list of tags is the same as passed parameter, 
previously we were putting it into PackageResult to make it easier to chain future calls, but now this method is used in different places so the method caller constructs PackageResult

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
andrii-paias: It's probably because initially it wasn't obvious in which situations tags can be updated, so POST/PUT requests have tags attribute, now that we know that tags are always updated independently from packages, we can probably remove tag update from POST and PUT,
I added a story for this - https://issues.folio.org/browse/MODKBEKBJ-283

sduvvuri-ebsco: perfect, thanks 👍

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
sduvvuri-ebsco: got it, thanks

PR #165, opened 2019-07-01
MODKBEKBJ-263 Add method to return list of Tags available for an entity type

stanislav-mureiko:
## Purpose
Providing new endpoint for list of unique tags.
GET /eholdings/tags/summary

## Approach
Adding new tags response schemes.
Providing functionality of returning unique tags from database.

### DIFF
--- a/ramls/examples/tags/unique_tags_get_200_response.json
+++ b/ramls/examples/tags/unique_tags_get_200_response.json
@@ -2,21 +2,21 @@
   "data": [
     {
       "id": "01276405-a07b-47d5-b224-742af3fe0387",
-      "type": "provider",
+      "type": "tags",
       "attributes": {
         "value": "title-tag"
       }
     },
     {
       "id": "52687cb3-ecd6-4570-a2e2-3e212f26bef8",
-      "type": "package",
+      "type": "tags",
       "attributes": {
         "value": "package-tag"
       }
     },
     {
       "id": "ed0fdcac-1292-4354-adc2-7b80b58638e3",
-      "type": "package",
+      "type": "tags",
       "attributes": {
         "value": "provider-tag2"
       }
--- a/ramls/examples/tags/unique_tags_get_200_response.json
+++ b/ramls/examples/tags/unique_tags_get_200_response.json
@@ -1,0 +1,30 @@
+{
+  "data": [
+    {
+      "name": {
+        "value": "package-tag1"
+      }
+    },
+    {
+      "name": {
+        "value": "provider-tag1"
+      }
+    },
+    {
+      "name": {
+        "value": "provider-tag2"
+      }
+    },
+    {
+      "name": {
+        "value": "package-tag2"
+      }
+    }
+  ],
+  "meta": {
+    "totalResults": 4
+  },
+  "jsonapi": {
+    "version": "1.0"
+  }
+}
--- a/ramls/examples/tags/unique_tags_get_200_response.json
+++ b/ramls/examples/tags/unique_tags_get_200_response.json
@@ -1,28 +1,29 @@
 {
   "data": [
     {
-      "name": {
-        "value": "package-tag1"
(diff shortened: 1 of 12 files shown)


ramls/examples/tags/unique_tags_get_200_response.json
```diff
@@ -1,28 +1,29 @@
 {
   "data": [
     {
-      "name": {
-        "value": "package-tag1"
+      "id": "01276405-a07b-47d5-b224-742af3fe0387",
+      "type": "provider",
```
carolegodfrey: Could the type on these all be the same - tags

src/test/java/org/folio/rest/impl/EholdingsTagsImplTest.java
```diff
@@ -158,34 +157,41 @@ public void shouldReturnEmptyUniqueTagsCollection() {
...
+    }
   }
 
-  private TagUniqueCollection toUniqueTagCollection(List<TagUniqueCollectionItem> tags){
-    return new TagUniqueCollection()
-      .withData(tags)
-      .withMeta(new MetaTotalResults().withTotalResults(tags.size()))
-      .withJsonapi(RestConstants.JSONAPI);
+  @Test
+  public void shouldReturnListOfUniqueTagsWithInvalidParams() {
```
carolegodfrey: Very minor - Should this be named `shouldReturnBadRequestWithInvalidParams`

src/main/java/org/folio/rest/converter/tags/TagsConverters.java
```diff
@@ -56,6 +57,8 @@ public TagCollection convert(@NonNull List<Tag> source) {
     @Override
     public TagUniqueCollectionItem convert(@NonNull String source) {
       return new TagUniqueCollectionItem()
+        .withId(UUID.randomUUID().toString())
+        .withType(RestConstants.TAGS_TYPE)
```
carolegodfrey: Type in code is set to tags -- earlier comment is just on the sample response

src/test/java/org/folio/rest/impl/EholdingsTagsImplTest.java
```diff
@@ -125,6 +130,70 @@ public void shouldFailOnInvalidRecordTypeOnGet() {
...
   }
 
+  @Test
+  public void shouldReturnAllUniqueTags() {
+    insertTags(UNIQUE_TAGS, vertx);
+
+    try {
+      TagUniqueCollection col = getWithOk("eholdings/tags/summary").as(TagUniqueCollection.class);
+
+      assertEquals(col.getData().size(), 4);
```
andrii-paias: It would be good if tests also checked that response contains correct tag values

ramls/tags.raml
```diff
@@ -56,3 +57,38 @@ types:
...
+          example:
+            - provider
+            - package
+      responses:
+        200:
+          description: OK
+          body:
+            application/vnd.api+json:
+              description: |
+                An array of tags comprising the results of the query sorted in alpahbetical order.
```
andrii-paias: typo in word alphabetical

ramls/types/tags/tagUniqueCollectionItem.json
```diff
@@ -0,0 +1,25 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Tag object schema for a collection",
+  "description": "Tag object schema for a collection of unique tags",
+  "javaType": "org.folio.rest.jaxrs.model.TagUniqueCollectionItem",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "UUID of tag assigned to the record",
```
andrii-paias: If unique tags don't really have any uuid associated with them, we should probably make it clear that this field will just contain a randomly generated uuid

PR #166, opened 2019-07-03
MODKBEKBJ-281 Add separate endpoint for updating provider tags

stanislav-mureiko:
## Purpose
1)Adding endpoint PUT /providers/{id}/tags with request body.
2)Implementing endpoint by using provider and tag data to update local db.
3)Adding API tests for PUT /providers/{id}/tags.

## Approach
Providing all steps for adding new endpoint(changing raml, adding schemes etc).
Adding unit tests.
Changing updateTags method for interaction with database.


src/test/java/org/folio/rest/impl/EholdingsProvidersImplTest.java
```diff
@@ -423,6 +425,68 @@ public void shouldUpdateOnlyProviderTagsWhenNoPackagesAreSelected() throws IOExc
...
+      List<String> tagsAfterRequest = TagsTestUtil.getTagsForRecordType(vertx, RecordType.PROVIDER);
+      assertThat(tagsAfterRequest, containsInAnyOrder(newTags.toArray()));
+    } finally {
+      TagsTestUtil.clearTags(vertx);
+      TestUtil.clearDataFromTable(vertx, PROVIDERS_TABLE_NAME);
+    }
+  }
+
+  @Test
+  public void shouldUpdateTagsOnPutTagsAndParams() throws IOException, URISyntaxException {
```
andrii-paias: because "?param" will be ignored, this test seems to be exactly the same as "shouldUpdateTagsOnPutTagsWithAlreadyExistingTags"

PR #175, opened 2019-08-06
MODKBEKBJ-283 - Remove update of package tags from package endpoints

natali-zaitseva:
## Purpose
updates for https://issues.folio.org/browse/MODKBEKBJ-283

## Approach
After discussion with UI team decided not to remove "tags" from the schemas for POST and PUT package methods as those values for tags are used for the sequential call to update package tags.

* removed tags update for POST /packages
* removed tags update for PUT /packages/{id}
* updated/removed redundant tests

### DIFF
--- a/ramls/types/packages/packagePostDataAttributes.json
+++ b/ramls/types/packages/packagePostDataAttributes.json
@@ -20,11 +20,6 @@
       "type": "object",
       "description": "Custom Coverage",
       "$ref": "../coverage.json"
-    },
-    "tags": {
-      "type": "object",
-      "description": "Package tags",
-      "$ref": "../../raml-util/schemas/tags.schema"
     }
   },
   "required": [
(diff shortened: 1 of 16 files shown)


natali-zaitseva:
UPD: agreed with UI team to return packages on POST and PUT requests without tags, it means that backend JSON schemas for packages can be cleaned up from "tags" property.

sduvvuri-ebsco:
Thanks Natalia. So, from your comment above, looks like we can cleanup tags schema from package post and put methods. I see that we cleaned up from POST schema, but in PUT I don't see tags schema at all. Has it been removed in a separate PR?

natali-zaitseva:
@sduvvuri-ebsco we used the same schema for the POST and PUT(I do not remember when we changed it) and in this PR I have separated schemas for POST and PUT requests, so now 
ramls/types/packages/packagePutData.json uses "packagePutDataAttributes.json" instead of "packageDataAttributes.json" which is used for POST requests.
Screen Shot 2019-08-30 at 5 06 25 PM

sduvvuri-ebsco:
ok, makes sense. Thanks Natalia.

PR #176, opened 2019-08-13
MODKBEKBJ-286 Remove update of provider tags from provider endpoints

stanislav-mureiko:
## Purpose
Remove code that updates tags https://issues.folio.org/browse/MODKBEKBJ-286

## Approach
* removed tags update for PUT /providers/{id}
* updated/removed redundant tests


stanislav-mureiko:
I have updated naming of "loadHoldings" due to duplication of schema names for "status".

PR #179, opened 2019-09-03
MODKBEKBJ-306: Correctly process "None" status on loading holdings

andrii-paias:
Jira: MODKBEKBJ-306

## Purpose
Correctly process "None" status on loading holdings

## Approach
Add None status

Additionally:
Fix an issue in WireMockTestBase to make it use TestConfig
Fix an issue in LoadHoldingsImplTest where HOLDINGS_GET_ENDPOINT returned
an error instead of HOLDINGS_STATUS_ENDPOINT

### DIFF
--- a/src/main/java/org/folio/repository/holdings/LoadStatus.java
+++ b/src/main/java/org/folio/repository/holdings/LoadStatus.java
@@ -4,7 +4,7 @@ import java.util.Arrays;
 
 public enum LoadStatus {
 
-  COMPLETED("Completed"), IN_PROGRESS("In progress"), FAILED("Failed");
+  COMPLETED("Completed"), IN_PROGRESS("In progress"), FAILED("Failed"), NONE("None");
 
   private String value;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/repository/holdings/LoadStatus.java
```diff
@@ -4,7 +4,7 @@
 
 public enum LoadStatus {
 
-  COMPLETED("Completed"), IN_PROGRESS("In progress"), FAILED("Failed");
+  COMPLETED("Completed"), IN_PROGRESS("In progress"), FAILED("Failed"), NONE("None");
```
sduvvuri-ebsco: If we get a status - NONE, then we should make a POST request to RM API for staging data. Are we handling that?

andrii-paias: Yes, we make POST request when we get any status except "In progress"

sduvvuri-ebsco: 👍

PR #181, opened 2019-09-06
MODKBEKBJ-312 - Update RMB to version 27.0.0

natali-zaitseva:
## Purpose
Due to https://issues.folio.org/browse/MODKBEKBJ-312 we want to be able to use new capabilities of the RMB module.

## Approach
* updated RMB to 27.0.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,7 +34,7 @@
   </licenses>
 
   <properties>
-    <raml-module-builder.version>24.0.0</raml-module-builder.version>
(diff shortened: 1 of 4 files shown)


natali-zaitseva:
Run API tests (based on [commit](https://github.com/folio-org/folio-api-tests/commit/755c81e52069ee4a9ae80ea8461c99b268c28071)) against this PR to double-check that nothing is broken and they completed successfully.

PR #193, opened 2019-11-21
MODKBEKBJ-320: Add audit table for holdings_status

andrii-paias:
## Purpose
Make is easier to track status of loading holdings for debugging purposes

## Approach
Create table holdings_status_audit
Add trigger to insert records into holdings_status_audit on every update
of holdings_status table
Periodically remove all records that are older than 30 days


src/main/resources/application.properties
```diff
@@ -7,6 +7,7 @@ title.cache.expire=86400
 configuration.base.url.default=https://sandbox.ebsco.io
 holdings.status.check.delay=120000
 holdings.status.retry.count=5
+holdings.status.audit.expiration.period=2592000000
```
dmtkachenko: is this in milliseconds?

andrii-paias: yes

PR #194, opened 2019-11-22
MODKBEKBJ-339: Update RMB version to 29.0.1

andrii-paias:
## Purpose
Update RMB to version 29.0.1

## Approach
Remove usage of .completer(), because .completer() return reference to this
Replace usages of Future with Promise
Fix other sonar issues


pom.xml
```diff
@@ -34,7 +34,7 @@
   </licenses>
 
   <properties>
-    <raml-module-builder.version>27.1.1</raml-module-builder.version>
+    <raml-module-builder.version>28.1.1</raml-module-builder.version>
     <vertx.version>3.8.1</vertx.version>
```
dmtkachenko: please remove this explicit dependency on vertx
vertx should be a transitive dependency that comes wtih RMB

andrii-paias: mod-kb-ebsco-java is using some modules that are not present in RMB, for example vertx-unit, vertx-service-proxy, also it is using vertx modules directly, so I think it is ok to have direct dependency on vertx

PR #196, opened 2019-11-29
MODKBEKBJ-333: Set fromModuleVersion attribute for all tables

andrii-paias:
## Purpose
Set fromModuleVersion attribute

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -2,36 +2,42 @@
   "tables": [
     {
       "tableName": "tags",
+      "fromModuleVersion": "mod-kb-ebsco-java-3.0.1",
       "withMetadata": false,
       "withAuditing": false,
       "customSnippetPath": "create-tags-table.sql"
     },
     {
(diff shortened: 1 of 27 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -61,12 +68,11 @@
     {
     "run": "after",
     "snippetPath": "update_started_date_before_insertion.sql",
-    "fromModuleVersion": "1.0"
+    "fromModuleVersion": "mod-kb-ebsco-java-3.0.1"
     },
     {
       "run": "after",
-      "snippetPath": "drop_set_id_in_jsonb.sql",
-      "fromModuleVersion": "mod-kb-ebsco-java-3.0.2"
```
andrii-paias: drop_set_id_in_jsonb.sql shouldn't have fromModuleVersion, because it has to run even if it wasn't changed (set_id_in_jsonb trigger can be added when running scripts for other tables, so we always need to run drop_set_id_in_jsonb.sql after all other scripts).

PR #208, opened 2020-01-28
MODKBEKBJ-346: Fix bug where resource is not removed if request has userDefinedFields

andrii-paias:
## Purpose
Make is possible to de-select resource if PUT /resources/{id} request contains userDefinedFields

## Approach
Allow empty userDefinedFields for PUT /resources/{id} requests where
isSelected=false.
Remove case in which PUT request is not sent to RM API if request is
invalid. This case was used to update tags when isSelected=false, but
now tags are updated in a separate endpoint.


andrii-paias:
@dmtkachenko Oh, sorry, forgot to commit tests

PR #210, opened 2020-02-04
MODKBEKBJ-348 -  Settings: CRUD Access Status Types

natali-zaitseva:
## Purpose
Per https://issues.folio.org/browse/MODKBEKBJ-348 add CRUD functionality to be able to add access types for eholdings

## Approach
* added endpoints description for CRUD operations
* added implementation for :
  ** GET /eholdings/access-types
  ** POST /eholdings/access-types
  ** GET /eholdings/access-types/{id}
  ** PUT /eholdings/access-types/{id}
  ** DELETE /eholdings/access-types/{id}
* covered new functionality with tests

For now, the restriction of the max number of access status types -  15 is hardcoded as it is the first version of the new API. Next, the max number can be made configurable if needed.


src/main/java/org/folio/service/userlookup/UserLookUpService.java
```diff
@@ -0,0 +1,99 @@
...
+    final String tenantId = TenantTool.calculateTenantId(headers.get(RestVerticle.OKAPI_HEADER_TENANT));
+    final String userId = headers.get(RestVerticle.OKAPI_USERID_HEADER);
+    CompletableFuture<UserLookUp> future = new CompletableFuture();
+    if (userId == null) {
+      logger.error("No userid header");
+      future.completeExceptionally(new BadRequestException("Missing user id header, cannot look up user"));
+      return future;
+    }
+
+    String okapiURL = headers.get("X-Okapi-Url");
```
psmagin:
```suggestion
    String okapiURL = headers.get(RestConstants.OKAPI_URL_HEADER);
```
better use constant

src/main/java/org/folio/service/userlookup/UserLookUpService.java
```diff
@@ -0,0 +1,99 @@
...
+    try {
+      final HttpClientInterface httpClient = HttpClientFactory.getHttpClient(okapiURL, tenantId);
+      httpClient.request(url, okapiHeaders)
+        .thenApply(response -> {
+          try {
+            if (Response.isSuccess(response.getCode())) {
+              return mapUserInfo(response);
+            } else if (response.getCode() == 401 || response.getCode() == 403) {
+              logger.error("Authorization failure");
+              throw new NotAuthorizedException("Authorization failure");
```
psmagin: Duplicate strings. Should be better to extract them to constants?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -85,6 +91,11 @@
       "run": "after",
       "snippetPath": "holdings_status_history_trigger.sql",
       "fromModuleVersion": "mod-kb-ebsco-java-3.0.2"
+    },
+    {
+      "run": "after",
+      "snippetPath": "access_types_update_metadata_user_info.sql",
+      "fromModuleVersion": "mod-kb-ebsco-java-3.0.1"
```
dmtkachenko: why does version number differ from the one used with _"tableName": "access_types",_

src/main/java/org/folio/rest/validator/ValidatorUtil.java
```diff
@@ -134,6 +141,17 @@ public static boolean isUrlValid(String url) {
     }
   }
 
+  public static boolean isUUIDValid(String uuid) {
```
dmtkachenko: I don't believe there is no any standard UUID validation. Why do we need to invent another one?

src/main/java/org/folio/service/userlookup/UserLookUp.java
```diff
@@ -0,0 +1,56 @@
...
+  private String firstName;
+  private String middleName;
+  private String lastName;
+
+  /**
+   * Returns the username.
+   *
+   * @return UserName.
+   */
+  public String getUserName() {
```
dmtkachenko: lombok can automatically generate getters.
any reason to explicitly defined them?

src/main/resources/templates/db_scripts/schema.json
natali-zaitseva: my bad, fixed

src/main/java/org/folio/service/userlookup/UserLookUpService.java
natali-zaitseva: created constants

src/main/java/org/folio/service/userlookup/UserLookUpService.java
natali-zaitseva: fixed

src/main/java/org/folio/service/userlookup/UserLookUp.java
natali-zaitseva: fixed

src/main/java/org/folio/rest/validator/ValidatorUtil.java
natali-zaitseva: standard UUID.fromString(id) does not correctly validates uuid(s). For instance the invalid value '99999999-9999-2-9999-999999999999' passes UUID.fromString(id) validation but it is not correct.

PR #211, opened 2020-02-06
MODKBEKBJ-349 Set access type on create custom package

psmagin:
## Purpose
On the Create Custom package assign it to access type if it exists in POST request 

## Approach
Created a new DB table where packages and access types are linking

### DIFF
--- a/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
+++ b/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
@@ -103,8 +103,6 @@ public class EholdingsPackagesImpl implements EholdingsPackages {
 
   private static final String INVALID_PACKAGE_TITLE = "Package cannot be deleted";
   private static final String INVALID_PACKAGE_DETAILS = "Invalid package";
-  private static final String ACCESS_TYPE_NOT_FOUND_MESSAGE = "Access type does not exist";
-  private static final String ACCESS_TYPE_NOT_FOUND_DETAILS_MESSAGE = "Access type with id %s does not exist.";
 
   @Autowired
   private PackageRequestConverter converter;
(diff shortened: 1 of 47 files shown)


psmagin:
Please review this and tell me if I am on the correct way to implement it. 
I will create tests after your review.

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -186,22 +189,55 @@ public void getEholdingsPackages(String filterCustom, String q, String filterSel
...
+      template.requestAction(context ->
+        accessTypesService.existsById(accessTypeId, okapiHeaders)
+          .thenCompose(aBoolean -> {
+            if (Boolean.TRUE.equals(aBoolean)) {
+              return getVendorId(context)
+                .thenCompose(id -> context.getPackagesService().postPackage(packagePost, id))
+                .thenApply(packageById -> new PackageResult(packageById, null, null))
+                .thenCompose(packageResult -> updateAccessType(accessTypeId, packageResult, okapiHeaders));
+            } else {
+              throw new ValidationException("Access type does not exist");
```
natali-zaitseva: Could you use InputValidationException instead of ValidationException, please?

natali-zaitseva: and use constant for an error message like ACCESS_TYPE_NOT_FOUND_MESSAGE as it probably will be used in the future PUT method

psmagin: Fixed.

psmagin:
Do we need to set up the access type in the `relationships` of the package as per jsonapi?

PR #218, opened 2020-02-26
MODKBEKBJ-354 Set access type on update resource

psmagin:
## Purpose
Implement story [MODKBEKBJ-354](https://issues.folio.org/browse/MODKBEKBJ-354).
Implement the possibility to set access status type when update resource.

## Approach
* Update ramls for resource
* Update `GET, PUT, DELETE` requests in `EholdingsResourcesImpl`
* Update `ResourceResultConverter` converter
* Make refactoring of `ResourcePutBodyValidator`
* Add new tests

### DIFF
--- a/src/main/java/org/folio/service/loader/RelatedEntitiesLoader.java
+++ b/src/main/java/org/folio/service/loader/RelatedEntitiesLoader.java
@@ -1,0 +1,16 @@
+package org.folio.service.loader;
+
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+import org.folio.repository.RecordKey;
+import org.folio.rmapi.result.Accessible;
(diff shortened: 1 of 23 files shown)


src/main/java/org/folio/service/loader/RelatedEntitiesLoader.java
```diff
@@ -0,0 +1,16 @@
+package org.folio.service.loader;
+
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+import org.folio.repository.RecordKey;
+import org.folio.rmapi.result.Accessible;
+import org.folio.rmapi.result.Tagable;
+
+public interface RelatedEntitiesLoader {
```
dmtkachenko: like the idea to have a special component responsible for loading types/tags

PR #219, opened 2020-03-02
MODKBEKBJ-378 add number of records to GET endpoints

psmagin:
## Purpose
Return number of records used by each Access Type ([JIRA](https://issues.folio.org/browse/MODKBEKBJ-378))

## Approach
* Update raml
* Update GET endpoints to return number of records used by each Access Type
* Update tests

### DIFF
--- a/ramls/types/accessTypes/accessTypeCollectionItem.json
+++ b/ramls/types/accessTypes/accessTypeCollectionItem.json
@@ -35,6 +35,11 @@
       "$ref": "userDisplayInfo.json",
       "readonly": true
     },
+    "recordUsage": {
+      "type": "integer",
+      "description": "Number of records that use the access type",
+      "readonly": true
+    },
     "metadata": {
       "type": "object",
       "description": "Metadata for the entity",
--- a/ramls/types/accessTypes/accessTypeCollectionItem.json
+++ b/ramls/types/accessTypes/accessTypeCollectionItem.json
@@ -35,7 +35,7 @@
       "$ref": "userDisplayInfo.json",
       "readonly": true
     },
-    "recordUsage": {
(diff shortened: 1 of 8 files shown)


ramls/types/accessTypes/accessTypeCollectionItem.json
```diff
@@ -35,6 +35,11 @@
       "$ref": "userDisplayInfo.json",
       "readonly": true
     },
+    "recordUsage": {
```
dmtkachenko: Could you rename it to usageNumber

src/main/java/org/folio/service/accesstypes/AccessTypesServiceImpl.java
```diff
@@ -59,12 +62,15 @@
   @Override
   public CompletableFuture<AccessTypeCollection> findAll(Map<String, String> okapiHeaders) {
     return repository.findAll(tenantId(okapiHeaders))
+      .thenCombine(mappingService.findAll(okapiHeaders), this::setEachRecordUsage)
```
dmtkachenko: This is not the best way of getting number of usages. When you have thousand of mappings the performance is going down significantly.
You need to perform the calculation on DB side by grouping records and counting the totals

ramls/types/accessTypes/accessTypeCollectionItem.json
psmagin: Renamed

src/main/java/org/folio/service/accesstypes/AccessTypesServiceImpl.java
psmagin: Fixed

PR #220, opened 2020-03-04
MODKBEKBJ-355 Implement package filtering by access type

psmagin:
## Purpose
Implement the ability to finding packages assigned to an access status type(s). [JIRA](https://issues.folio.org/browse/MODKBEKBJ-355)

## Approach
* Update raml
* Implement filtering
  * Implement finding `access status type` by its `name`
  * Implement finding `access status type mappings` by `access status type ids` and `RecordType`
* Create new tests


dcrossleyau:
Note that there is a spelling error with the filename: ramls/traits/accesible.raml
Should be "accessible". Better to fix now, as it will bite people later.

PR #225, opened 2020-03-11
MODKBEKBJ-375 Implement filtering packages by access type and provider

psmagin:
## Purpose
Implement filtering packages with a provider assigned to an access status type(s) 

## Approach
* Update `providers.raml`
* Implement feature in `EholdingsProvidersImpl` 
* Extract loading packages to new `FilteredEntitiesLoader`
* Make refactoring of `IdParser` and `AccessTypeMappingsRepository`
* Add tests


src/main/java/org/folio/rest/model/filter/AccessTypeFilter.java
```diff
@@ -0,0 +1,22 @@
...
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+import org.folio.repository.RecordType;
+
+@Getter
+@Setter
+@Builder
```
dmtkachenko: why do you need builder if you allow setters?

usually builder is used to create object and populate its properties, the object itself remains unchangeable once created
another potential use case of builder is to create objects with complex internal structure or if creation process is so complex that it's reasonable to encapsulate the process into a specific class (builder)

psmagin: Thank you for the clear explanation.

PR #231, opened 2020-03-25
MODKBEKBJ-384 Apply sort to packages

psmagin:
## Purpose
Apply sorting to the list of packages in title record in this order
1. Packages Starting with Special Chars (Etc #,&,(….)
2. Packages Starting with Numbers
3. Packages Starting with English names (Sorted by A-Z)
4. Packages Starting with Foreign Characters.
5. Packages without Package Names

## Approach
* update `TitleConverter` with sorting by package name
* update unit tests (`pom.xml` was changed to have ability testing not-Latin test data)

### DIFF
--- a/src/main/java/org/folio/rest/converter/titles/TitleConverter.java
+++ b/src/main/java/org/folio/rest/converter/titles/TitleConverter.java
@@ -4,11 +4,13 @@ import static org.folio.common.ListUtils.mapItems;
 import static org.folio.rest.converter.titles.TitleConverterUtils.createEmptyResourcesRelationships;
 import static org.folio.rest.util.RestConstants.TITLES_TYPE;
 
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.Iterator;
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/rest/converter/titles/TitleConverter.java
```diff
@@ -100,10 +103,34 @@
     return title;
   }
 
+  private List<CustomerResources> getSortedCustomerResourcesList(List<CustomerResources> customerResourcesList) {
+    if (customerResourcesList != null) {
```
natali-zaitseva: do we have to check customerResourcesList != null twice? https://github.com/folio-org/mod-kb-ebsco-java/blob/5ad25779e56d2edcf786959ffdcbe5481b3415ee/src/main/java/org/folio/rest/converter/titles/TitleConverter.java#L82

psmagin: Agree

PR #235, opened 2020-04-03
MODKBEKBJ-420 Implement endpoint to create KB credentials

psmagin:
## Purpose
Support `POST /eholdings/kb-credentials` to retrieve KB credentials list

## Approach
* Add raml definitions
* Update `ModuleDescriptor`
* Implement endpoint to save KB credentials
* Add tests


src/test/resources/test-application.properties
```diff
@@ -1,26 +1,34 @@
...
 holdings.timeout=60000
-# Qualifier of a bean that implements LoadServiceFacade interface.
-# Possible values are TransactionLoadServiceFacade, DefaultLoadServiceFacade
-holdings.load.implementation.qualifier=DefaultLoadServiceFacade
-access.types.default.number.limit.value=3
-db.exception.translator.name=postgresql
+
+# Credentials properties
+kb.ebsco.credentials.name.length.max=255
+kb.ebsco.credentials.url.default=https://sandbox.ebsco.io
```
carolegodfrey, line 24: Really minor comment -- could this default to https://api.ebsco.com? This will be the default/used most of the time

psmagin: Yes, I will change it.

src/main/resources/application.properties
```diff
@@ -1,26 +1,34 @@
...
 holdings.timeout=3600000
-# Qualifier of a bean that implements LoadServiceFacade interface.
-# Possible values are TransactionLoadServiceFacade, DefaultLoadServiceFacade
-holdings.load.implementation.qualifier=DefaultLoadServiceFacade
-access.types.default.number.limit.value=15
-db.exception.translator.name=postgresql
+
+# Credentials properties
+kb.ebsco.credentials.name.length.max=255
+kb.ebsco.credentials.url.default=https://sandbox.ebsco.io
```
carolegodfrey, line 24: My comment regarding using https://api.ebsco.com -- applies to these properties instead of test properties

ramls/examples/kbCredentials/kb_credentials_post_201_response.json
```diff
@@ -0,0 +1,18 @@
+{
+  "id": "2ffa1940-2cf6-48b1-8cc9-5e539c61d93f",
+  "type": "kbCredentials",
+  "attributes": {
+    "name": "University of Massachusetts",
+    "apiKey": "XXXX",
```
carolegodfrey, line 6: Just want to confirm that the actual apiKey does not get returned in the POST response (we had **** it in the past)

src/main/java/org/folio/repository/kbcredentials/KbCredentialsRepositoryImpl.java
```diff
@@ -56,6 +64,29 @@
     return mapResult(promise.future().recover(excTranslator.translateOrPassBy()), this::mapCredentialsCollection);
   }
 
+  @Override
+  public CompletableFuture<DbKbCredentials> save(DbKbCredentials credentials, String tenant) {
+    String query = format(INSERT_CREDENTIALS_QUERY, getKbCredentialsTableName(tenant));
+
+    String id = UUID.randomUUID().toString();
```
dmtkachenko, line 71: image
It's a common pattern on Folio to allow id to be passed outside (in the request) for a new record.
If it's not present then generate it

src/test/resources/test-application.properties
psmagin: Changed

src/main/resources/application.properties
psmagin: Changed too.

ramls/examples/kbCredentials/kb_credentials_post_201_response.json
psmagin: Update example with ****. Implementation will return the ***** API key.

src/main/java/org/folio/repository/kbcredentials/KbCredentialsRepositoryImpl.java
psmagin: Fixed

PR #239, opened 2020-04-07
MODKBEKBJ-422 Implement endpoint to update credentials by ID

psmagin:
## Purpose
Support `PUT  /eholdings/kb-credentials{credentialsId}` to update existing KB credentials entry

## Approach
* Add raml definitions
* Update interface in ModuleDescriptor
* Implement endpoint to update credentials by ID
* Add tests


src/main/java/org/folio/rest/validator/KbCredentialsPutBodyValidator.java
```diff
@@ -0,0 +1,24 @@
...
+  private final int nameLengthMax;
+
+  public KbCredentialsPutBodyValidator(@Value("${kb.ebsco.credentials.name.length.max:255}") int nameLengthMax) {
+    this.nameLengthMax = nameLengthMax;
+  }
+
+  public void validate(KbCredentialsPutRequest entity) {
+    String name = entity.getData().getAttributes().getName();
+    ValidatorUtil.checkIsNotBlank(NAME_PARAMETER, name);
+    ValidatorUtil.checkMaxLength(NAME_PARAMETER, name, nameLengthMax);
```
dmtkachenko, line 22: please validate other fields that are known to be mandatory, like apiKey etc
and do the same in POST validator

PR #243, opened 2020-04-13
MODKBEKBJ-408 - Support GET /eholdings/kb-credentials/{credId}/users to retrieve users associated with KBs

natali-zaitseva:
## Purpose
Implementation for https://issues.folio.org/browse/MODKBEKBJ-408

## Approach
* added a separate raml file for assigned-users functionality
* added an implementation for GET /eholdings/kb-credentials/{credId}/users
* added tests

## TODOS and Open Questions
- [ ] the correctness of URI parameter {credId} will be handled on DB side.
- [ ] Postman API tests will be added in next stories related to user assignment


ramls/types/assignedUsers/assignedUserDataAttributes.json
```diff
@@ -0,0 +1,45 @@
...
+      "examples": "Doe"
+    },
+    "patronGroup": {
+      "type": "string",
+      "description": "The patron group name",
+      "examples": "Staff"
+    }
+  },
+  "required": [
+    "credentialsId",
```
psmagin, line 42: I think all fields except `middleName` are required because they will be used in the frontend.

src/main/resources/templates/db_scripts/create-assigned-users-table.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE assigned_users
+DROP COLUMN IF EXISTS jsonb,
+ADD COLUMN IF NOT EXISTS credentials_id uuid NOT NULL constraint fk_assigned_users_kb_credentials references kb_credentials,
```
psmagin, line 3: Does this type of constraint declaration create new constraint each time DB is created?

ramls/types/assignedUsers/assignedUserDataAttributes.json
natali-zaitseva: well, on user creation page 
Screenshot 2020-04-13 at 17 53 18
there are several fields are required - Last name, patron group, status and email. A user can be created even without user name(but can not login into the system), middle name and first name.
so, the middle name is not required but thank you for review, because I found inaccuracies in JSON schemas.

psmagin: You add firstName but should lastName according to the info above

psmagin: Also db schema must be changed to have nullable columns for not required fields

natali-zaitseva: thanks, fixed

src/main/java/org/folio/rest/impl/EholdingsAssignedUsersImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+
+import org.folio.rest.annotations.Validate;
+import org.folio.rest.aspect.HandleValidationErrors;
+import org.folio.rest.jaxrs.resource.EholdingsKbCredentialsId;
+import org.folio.rest.util.ErrorHandler;
+import org.folio.service.assignedusers.AssignedUsersService;
+import org.folio.spring.SpringContextUtil;
+
+public class EholdingsAssignedUsersImpl implements EholdingsKbCredentialsId {
```
dmtkachenko, line 23: `EholdingsKbCredentialsId` is not a good name. Any way to change it?

natali-zaitseva: Changed the name, thanks for review

PR #247, opened 2020-04-15
MODKBEKBJ-412 - Support DELETE /eholdings/kb-credentials/{credId}/users/{userId} to remove association between user and KB

natali-zaitseva:
## Purpose
implementation story for https://issues.folio.org/browse/MODKBEKBJ-412

## Approach
* modified ModuleDescriptor file
* modified assigned-users.raml file
* added implementation for DELETE /eholdings/kb-credentials/{credId}/users/{userId}
* added tests

### DIFF
--- a/src/main/java/org/folio/repository/assigneduser/AssignedUserRepositoryImpl.java
+++ b/src/main/java/org/folio/repository/assigneduser/AssignedUserRepositoryImpl.java
@@ -43,7 +43,6 @@ import org.springframework.stereotype.Component;
 import org.folio.db.exc.ConstraintViolationException;
 import org.folio.db.exc.DbExcUtils;
 import org.folio.db.exc.translation.DBExceptionTranslator;
-import org.folio.rest.jaxrs.model.AssignedUser;
 import org.folio.rest.persist.PostgresClient;
 import org.folio.service.exc.ServiceExceptions;
 
@@ -117,12 +116,9 @@ public class AssignedUserRepositoryImpl implements AssignedUserRepository {
     Promise<UpdateResult> promise = Promise.promise();
     pgClient(tenant).execute(query, params, promise);
 
-    return mapResult(promise.future().recover(excTranslator.translateOrPassBy()), updateResult -> {
-      if (updateResult.getUpdated() == 0) {
-        throw ServiceExceptions.notFound(AssignedUser.class, dbAssignedUser.getId());
-      }
-      return null;
-    });
+    Future<UpdateResult> resultFuture = promise.future()
+      .recover(excTranslator.translateOrPassBy());
+    return mapResult(resultFuture, updateResult -> checkUserFound(dbAssignedUser.getId(), updateResult));
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/repository/assigneduser/AssignedUsersConstants.java
```diff
@@ -30,11 +36,13 @@
     String[] updateColumns = new String[] {
       USER_NAME, FIRST_NAME, MIDDLE_NAME, LAST_NAME, PATRON_GROUP
     };
+    ASSIGNED_USER_COLUMN_LIST = String.format(join(",", nCopies(7, "%s")), (Object[]) allColumns);
```
psmagin, line 39: Don't see where this const is used.

src/test/java/org/folio/rest/impl/integrationsuite/EholdingsAssignedUsersImplTest.java
```diff
@@ -152,56 +156,40 @@ public void shouldReturn422OnPostWhenAssignedUserDoesNotHaveRequiredParameters()
...
+    String endpoint = String.format(ASSIGN_USER_PATH, credentialsId);
+    Errors errors = postWithStatus(endpoint, postBody, SC_UNPROCESSABLE_ENTITY).as(Errors.class);
 
     assertThat(errors.getErrors(), hasSize(5));
     assertThat(errors.getErrors(), everyItem(hasProperty("message", equalTo("may not be null"))));
   }
 
   @Test
-  public void shouldReturn204OnPutWhenAssignedUserIsValid() {
+  public void shouldReturn204OnDeleteUserAssignment(){
```
psmagin, line 167: This tests for `put` are needed.

src/test/java/org/folio/util/AssignedUsersTestUtil.java
```diff
@@ -64,14 +61,6 @@ public static String insertAssignedUser(String credentialsId, String username, S
     return future.join();
   }
 
-  public static String resourcePath(String... params) {
```
psmagin, line 67: Why do you think that this method is not useful?

src/main/java/org/folio/repository/assigneduser/AssignedUserRepositoryImpl.java
```diff
@@ -122,6 +125,27 @@
...
+    Promise<UpdateResult> promise = Promise.promise();
+    pgClient(tenant).execute(query, createParams(Arrays.asList(credentialsId, userId)), promise);
+
+    Future<UpdateResult> resultFuture = promise.future()
+      .recover(excTranslator.translateOrPassBy());
+    return mapResult(resultFuture, updateResult -> checkUserDeleted(userId, updateResult));
+  }
+
+  @Nullable
+  private Void checkUserDeleted(String userId, UpdateResult updateResult) {
```
psmagin, line 142: This method also can be used in `update` method

src/main/java/org/folio/repository/assigneduser/AssignedUsersConstants.java
natali-zaitseva: removed

src/test/java/org/folio/rest/impl/integrationsuite/EholdingsAssignedUsersImplTest.java
natali-zaitseva: probably during the rebase I accidentally removed them. Thank you for note, added them back

src/test/java/org/folio/util/AssignedUsersTestUtil.java
natali-zaitseva: it is not clear from my perspective what is the resource path in the current test is used and the method name does not help to understand the context of  resourcePath

src/main/java/org/folio/repository/assigneduser/AssignedUserRepositoryImpl.java
natali-zaitseva: added to `update` method

PR #252, opened 2020-04-22
MODKBEKBJ-402 Make get access types to support multiple KB credentials

psmagin:
## Purpose
* Update `GET /eholdings/access-types` to support multiple KB credentials
* Add `GET /eholdings/kb-credentials/<credentialsId>/access-types`  to support multiple KB credentials

## Approach
* Update Module descriptor and raml definitions
* Update `GET /eholdings/access-types` endpoint implementation to fetch access types that belong to current user credentials
 * Implement `GET /eholdings/kb-credentials/<credentialsId>/access-types` to fetch access types that belongs to specific credentials
* Update tests

#### TODOS and Open Questions
- [ ] a new DB table, repository, and service were created to left all staff work properly.  After everything related to access types will be implemented - they need to be deleted.


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -96,6 +89,20 @@
...
+    {
+      "tableName": "access_types",
+      "fromModuleVersion": "mod-kb-ebsco-java-3.4.1",
+      "withMetadata": false,
+      "withAuditing": false,
+      "customSnippetPath": "create-access-types-table.sql"
+    },
+    {
+      "tableName": "access_types_mapping",
+      "fromModuleVersion": "mod-kb-ebsco-java-3.3.0",
```
natali-zaitseva, line 102: shouldn't it be the same version as for "access_types" table?

src/main/resources/templates/db_scripts/create-access-types-table.sql
```diff
@@ -0,0 +1,17 @@
+ALTER TABLE access_types
+DROP COLUMN IF EXISTS jsonb,
+ADD COLUMN IF NOT EXISTS credentials_id UUID NOT NULL CONSTRAINT access_types_kb_credentials_fkey REFERENCES kb_credentials,
+ADD COLUMN IF NOT EXISTS name VARCHAR (100) NOT NULL CONSTRAINT name_unique UNIQUE,
```
natali-zaitseva, line 4: we have a validation on name length - kb.ebsco.credentials.access.types.name.length.max=75

src/main/resources/templates/db_scripts/create-access-types-table.sql
```diff
@@ -0,0 +1,17 @@
...
+ADD COLUMN IF NOT EXISTS description VARCHAR (200),
```
natali-zaitseva, line 5: same validation on description length - kb.ebsco.credentials.access.types.description.length.max=150

src/main/java/org/folio/rest/validator/AccessTypesBodyValidator.java
```diff
@@ -12,19 +13,28 @@
   private static final String INVALID_REQUEST_BODY_TITLE = "Invalid request body";
   private static final String INVALID_REQUEST_BODY_DETAILS = "Json body must contain data.attributes";
 
+  private final int maxNameLength;
+  private final int maxDescriptionLength;
+
+  public AccessTypesBodyValidator(@Value("${kb.ebsco.credentials.access.types.name.length.max:75}") int maxNameLength,
+                                  @Value("${kb.ebsco.credentials.access.types.description.length.max:150}") int maxDescriptionLength) {
+    this.maxNameLength = maxNameLength;
```
natali-zaitseva, line 21: the max value for name and description length specified in both application.properties and test-application.properties. there is no need to specify default value for them.

src/main/resources/templates/db_scripts/create-access-types-table.sql
psmagin: But this setup can be changed in properties. I decide to add some buffer to have a possibility to increase length without changing a DB schema.

src/main/java/org/folio/rest/validator/AccessTypesBodyValidator.java
psmagin: If smbd delete this property from properties-file than the application will work with these default values. I think it is a plus to fault tolerance.

PR #264, opened 2020-05-06
MODKBEKBJ-432  Modify configuration routine to support user related KB credentials

dmtkachenko:
## Purpose
Modify configuration routine to support user related KB credentials

## Approach
- add new `UserKbCredentialsService` to encapsulate credential search for a user   
- override `ConfigurationServiceImpl.retrieveConfiguration()` to get credentials from module's DB
- remove `EholdingsConfigurationImpl` as it becomes obsolete
- updates test to use credentials from DB instead of mocking the call to mod-config


ramls/eholdings.raml
```diff
@@ -11,63 +11,10 @@ documentation:
 
 types:
   configuration: !include types/configuration/configuration.json
-  configurationPutRequest: !include types/configuration/configurationPutRequest.json
   jsonapiError: !include types/jsonapiError.json
```
psmagin, line 14: `configuration` and `jsonapiError` types are not used in this raml file

dmtkachenko: I left them just to make sure that java classes generated

PR #294, opened 2020-07-13
MODKBEKBJ-435 Update DB schema to have separate user table

psmagin:
## Purpose
Update DB schema to have separate user table

## Approach
* create users table
* update existed tables anc create migration scripts
* create access type view
* make changes in implementation
* update tests


src/main/resources/liquibase/tenant/scripts/v3.5.3/alter-access-types-table.xml
```diff
@@ -0,0 +1,45 @@
...
+    <dropColumn tableName="access_types" columnName="created_by_first_name"/>
+    <dropColumn tableName="access_types" columnName="created_by_middle_name"/>
+    <dropColumn tableName="access_types" columnName="updated_by_username"/>
+    <dropColumn tableName="access_types" columnName="updated_by_last_name"/>
+    <dropColumn tableName="access_types" columnName="updated_by_first_name"/>
+    <dropColumn tableName="access_types" columnName="updated_by_middle_name"/>
+  </changeSet>
+
+  <changeSet id="MODKBEKBJ-435@@add-foreign-keys-from-access-types-to-kb-users" author="psmagin">
+    <addForeignKeyConstraint
```
dmtkachenko, line 19: it's better to have statements separately. thus they can be applied independently: if one fails it can be reapplied later because it belongs to different changeset
see https://www.liquibase.org/get-started/best-practices
image

PR #313, opened 2020-10-21
MODKBEKBJ-499 Implement package cost-per-use endpoint

psmagin:
## Purpose
Implement package cost-per-use endpoint

## Approach
* create raml definitions
* create new method int apigee client for fetching package cost-per-use
* implement endpoint
* using cache for multiply title-packages cost-per-use
* create tests


dmtkachenko (approved):

dmtkachenko (commented):
Looks good. 
The only thing, I'd consider making UC title cache active/inactive via application properties and deactivate it initially. We cannot estimate the amount of data that might be cached potentially, I assume it can be significant and might lead to memory issues with insufficient memory allocated

PR #325, opened 2020-11-06
MODKBEKBJ-487 Implement resources' cost-per-use sort

psmagin:
## Purpose
Make `/eholdings/packages/<packageId>/resources/costperuse` sortable.

## Approach
* Add new parameters to endoint:
  * `sort` with possible values:
    * name
    * type
    * cost
    * usage
    * costperuse
    * percent
  * `order` with possible value:
    * asc
    * desc
* Implement calculating `percent of cost` (it was accidentally missed in previous PR)
* Actions have been taken to improve performance:
  * add index to `package_id` in `holdings` table
  * add cache for KB Credentials

#### TODOS and Open Questions
- [ ] The performance of work with the `holdings` table can be improved by changing the type of the `package_id` column to an integer. Also, the `vendor_id` and `title_id` columns could be migrated.

**varchar column type query analysis:**
with_simple_index

**integer column type query analysis:**
with_simple_index_integer

### DIFF
--- a/ramls/costperuse.raml
+++ b/ramls/costperuse.raml
@@ -12,6 +12,8 @@ documentation:
 traits:
   costPerUseCommon: !include traits/costPerUseCommon.raml
   pageable: !include traits/pageable.raml
+  sortable: !include traits/sortable.raml
+  orderable: !include traits/orderable.raml
 
 types:
   resourceCostPerUse: !include types/uc/costperuse/resourceCostPerUse.json
@@ -19,6 +21,8 @@ types:
   packageCostPerUse: !include types/uc/costperuse/packageCostPerUse.json
   titleCostPerUse: !include types/uc/costperuse/titleCostPerUse.json
   jsonapiError: !include types/jsonapiError.json
+  order:
+    enum: [asc, desc]
 
 /eholdings/resources/{resourceId}/costperuse:
   displayName: Cost-per-use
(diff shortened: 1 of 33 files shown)


ramls/costperuse.raml
```diff
@@ -141,7 +145,9 @@ types:
       example: "15-440285"
   get:
     description: Retrieve cost-per-use information for package resources
-    is: [costPerUseCommon, pageable: {maxCountValue: 1000, defaultCountValue: 100}]
```
dmtkachenko, line 144: :thumbsup:

ramls/costperuse.raml
```diff
@@ -141,7 +145,9 @@ types:
...
+    is: [costPerUseCommon, orderable,
+    sortable: {defaultValue: 'name', possibleValues: 'name, type, cost, usage, costperuse, percent'},
+    pageable: {maxCountValue: 1000, defaultCountValue: 100}]
```
dmtkachenko, line 150: :thumbsup:

src/main/java/org/folio/service/kbcredentials/UserKbCredentialsServiceImpl.java
```diff
@@ -35,23 +37,18 @@
...
 
-
-  public UserKbCredentialsServiceImpl(KbCredentialsRepository credentialsRepository,
-                                      AssignedUserRepository assignedUserRepository,
-                                      Converter<DbKbCredentials, KbCredentials> credentialsFromDBConverter) {
-    this.credentialsRepository = credentialsRepository;
-    this.assignedUserRepository = assignedUserRepository;
-    this.credentialsFromDBConverter = credentialsFromDBConverter;
-  }
+  private final VertxCache<String, KbCredentials> cache;
```
dmtkachenko, line 40: Let's think carefully if it's possible to avoid using cache. The problem might be with excessive number of calls to the service but not with the long response.

dmtkachenko (approved):

natali-zaitseva (approved):

PR #326, opened 2020-11-09
MODKBEKBJ-523 Implement /eholdings/uc endpoint

psmagin:
## Purpose
Implement `/eholdings/uc` endpoint

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -727,6 +727,7 @@
       "displayName": "EBSCO KB Broker Usage Consolidation settings - all permissions",
       "description": "All permissions for managing Usage Consolidation settings",
       "subPermissions": [
+        "kb-ebsco.uc.item.get",
         "kb-ebsco.kb-credentials.uc.item.get",
         "kb-ebsco.kb-credentials.uc.collection.post",
         "kb-ebsco.kb-credentials.uc.item.patch",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -326,6 +326,11 @@
         },
         {
           "methods": ["GET"],
+          "pathPattern": "/eholdings/uc",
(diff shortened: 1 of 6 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -677,6 +682,11 @@
         "kb-ebsco.kb-credentials.access-types.item.delete"
       ]
     },
+    {
```
natali-zaitseva, line 685: will this permission be included in some of the "all" permission set?

src/main/java/org/folio/rest/impl/EholdingsUsageConsolidationImpl.java
```diff
@@ -23,39 +23,39 @@
...
 
   @Override
-  public void getEholdingsKbCredentialsUcById(String credentialsId, Map<String, String> okapiHeaders,
-                                              Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
-    settingsService.fetchByCredentialsId(credentialsId, okapiHeaders)
+  public void getEholdingsUc(Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
+                             Context vertxContext) {
+    settingsService.fetchByUser(okapiHeaders)
       .thenAccept(ucSettings -> asyncResultHandler.handle(succeededFuture(
         GetEholdingsKbCredentialsUcByIdResponse.respond200WithApplicationVndApiJson(ucSettings))))
```
natali-zaitseva, line 49: Different response class. should be GetEholdingsUcResponse

descriptors/ModuleDescriptor-template.json
psmagin: Thanks! Fixed

src/main/java/org/folio/rest/impl/EholdingsUsageConsolidationImpl.java
psmagin: Fixed

natali-zaitseva (approved):

PR #350, opened 2021-01-19
MODKBEKBJ-561 Create endpoint that return decrypted KB key

psmagin:
## Purpose
Create endpoint that returns decrypted KB key

## Approach
New endpoint was implemented `/eholdings/kb-credentials/<id>/key`

### DIFF
--- a/ramls/kb-credentials.raml
+++ b/ramls/kb-credentials.raml
@@ -256,7 +256,7 @@ types:
               example: "Internal server error, contact administrator"
     /key:
       get:
-        description: Get a specific KB credentials by id.
+        description: Get a specific KB credentials key by id.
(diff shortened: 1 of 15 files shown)


dmtkachenko (approved):

ramls/kb-credentials.raml
```diff
@@ -253,6 +254,34 @@ types:
           body:
             text/plain:
               example: "Internal server error, contact administrator"
+    /key:
+      get:
+        description: Get a specific KB credentials by id.
```
natali-zaitseva, line 259: Get specific KB credentials key by id.

natali-zaitseva (approved):

PR #355, opened 2021-02-24
MODKBEKBJ-551 Upgrade to RMB v32.1 and Vert.X v4

psmagin:


pom.xml
```diff
@@ -34,31 +34,31 @@
...
-    <rmb.version>31.1.5</rmb.version>
-    <folio-di-support.version>1.1.0</folio-di-support.version>
-    <folio-service-tools.version>1.6.0</folio-service-tools.version>
-    <folio-holdingsiq-client.version>1.10.3</folio-holdingsiq-client.version>
-    <folio-liquibase-util.version>1.0.0</folio-liquibase-util.version>
-    <mod-configuration-client.version>4.0.3</mod-configuration-client.version>
+    <rmb.version>32.1.0</rmb.version>
+    <folio-di-support.version>1.3.0</folio-di-support.version>
+    <folio-service-tools.version>1.6.1</folio-service-tools.version>
+    <folio-holdingsiq-client.version>1.10.4-SNAPSHOT</folio-holdingsiq-client.version>
```
sduvvuri-ebsco, line 40: Is my understanding correct that snapshot was for testing purposes only?

pom.xml
```diff
@@ -34,31 +34,31 @@
...
+    <folio-liquibase-util.version>1.2.0</folio-liquibase-util.version>
+    <mod-configuration-client.version>5.6.0</mod-configuration-client.version>
```
sduvvuri-ebsco, line 42: Do we still use mod-configuration client? I remember at one point, we used that as a secret store but we no longer use it - right?

pom.xml
psmagin: The version will be changed while the releasing procedure.

pom.xml
psmagin: We use it to get locale from mod-configuration also.

natali-zaitseva: Pavlo is right, we are fetching the locale to format number values based on it.

natali-zaitseva (approved):

PR #364, opened 2021-05-26
MODKBEKBJ-581 Upgrade to RMB v33

psmagin:
## Purpose
Upgrade to RMB v33
https://issues.folio.org/browse/MODKBEKBJ-581


src/test/java/org/folio/service/users/UsersLookUpServiceTest.java
```diff
@@ -145,24 +143,6 @@ public void shouldReturn404WhenUserNotFound(TestContext context) {
     });
   }
 
-  @Test
-  public void shouldReturn500WhenMissingOkapiURLHeader(TestContext context) {
```
psmagin, line 149: The test was removed because the check for existed okapiUrl header performed in OkapiParams. So there is no chance that okapiParams comes to the service without that header.

oleksandr-dekin (approved):

natali-zaitseva (approved):

PR #372, opened 2021-10-06
MODKBEKBJ-596: Packages, error message when "isCustom" not provided

Shans-Kaluhin:
## Purpose
Endpoint PUT eholding/packages/{id}
If "isCustom" field is missing, server responds with error message like
"Attribute PackageName is missing."

The error message is not valid and entity doesn't even have an attribute "packageName".
So we need to improve error message.

## Approach
- If package "isCustom" not matched with origin package. Than you can see message like this
image

## Learning
[MODKBEKBJ-596](https://issues.folio.org/browse/MODKBEKBJ-596)

### DIFF
--- a/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
+++ b/src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
@@ -238,18 +238,18 @@ public class EholdingsPackagesImpl implements EholdingsPackages {
                                               Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     PackageId parsedPackageId = parsePackageId(packageId);
     templateFactory.createTemplate(okapiHeaders, asyncResultHandler)
-      .requestAction(context -> {
-        validateIsCustomMatch(entity, parsedPackageId, context);
-        return fetchAccessType(entity, context)
-          .thenCompose(accessType -> processUpdateRequest(entity, parsedPackageId, context)
+      .requestAction(context -> context.getPackagesService().retrievePackage(parsedPackageId)
+        .thenCompose(packageByIdData -> fetchAccessType(entity, context)
+          .thenCompose(accessType -> processUpdateRequest(entity, packageByIdData, context)
             .thenCompose(o -> {
               CompletableFuture<PackageByIdData> future = context.getPackagesService().retrievePackage(parsedPackageId);
               return handleDeletedPackage(future, parsedPackageId, context);
             })
             .thenApply(packageById -> new PackageResult(packageById, null, null))
             .thenCompose(packageResult -> updateAccessTypeMapping(accessType, packageResult, context))
-          );
-      })
+          )
+        )
+      )
       .addErrorMapper(NotFoundException.class, error400NotFoundMapper())
       .addErrorMapper(InputValidationException.class, error422InputValidationMapper())
       .executeWithResult(Package.class);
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/rest/validator/PackagePutBodyValidator.java
```diff
@@ -34,11 +34,7 @@ public void validate(PackagePutRequest request) {
     String value = attributes.getPackageToken() != null ? attributes.getPackageToken().getValue() : null;
 
     if (isSelected == null || !isSelected) {
-      ValidatorUtil.checkFalseOrNull("allowKbToAddTitles", allowKbToAddTitles);
```
Shans-Kaluhin, line 37: Deleted all this checks (and tests for it), because we can never reach this case. 
We were throwing an error when isCustom and isSelected both are false. 
So I remove this checks and transferred throwing here

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -505,20 +500,14 @@ private String getResourceId(TitleResult titleResult) {
       packagePutBodyValidator.validate(entity);
       packagePutBody = converter.convertToRMAPIPackagePutRequest(entity);
     }
-    return context.getPackagesService().updatePackage(parsedPackageId, packagePutBody);
+    return context.getPackagesService().updatePackage(parsePackageId(originalPackage.getFullPackageId()), packagePutBody);
   }
 
-  private boolean isPackageUpdatable(PackagePutRequest entity) {
```
psmagin, line 511: I think we should have this validation also. Because we can't update not selected packages that are not custom.

src/main/java/org/folio/rest/validator/PackagePutBodyValidator.java
```diff
@@ -34,11 +34,7 @@ public void validate(PackagePutRequest request) {
...
-      ValidatorUtil.checkFalseOrNull("visibilityData.isHidden", isHidden);
-      ValidatorUtil.checkIsEmpty("customCoverage.beginCoverage", beginCoverage);
-      ValidatorUtil.checkIsEmpty("customCoverage.endCoverage", endCoverage);
-      ValidatorUtil.checkIsNull("packageToken.value", value);
```
psmagin, line 41: Why did you remove all this validations?

psmagin (changes requested):

src/main/java/org/folio/rest/validator/PackagePutBodyValidator.java
Shans-Kaluhin: I will open another task for this

psmagin (approved):

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -521,6 +523,13 @@ private boolean isPackageUpdatable(PackagePutRequest entity) {
     return true;
   }
 
+  private void validateIsCustomMatch(PackageByIdData originalPackage, PackagePutRequest entity) {
+    if (!originalPackage.getIsCustom().equals(entity.getData().getAttributes().getIsCustom())) {
```
oleksandr-dekin, line 527: Could we move logic in separate method for better visibility?

src/main/java/org/folio/rest/impl/EholdingsPackagesImpl.java
```diff
@@ -243,14 +244,16 @@ public void putEholdingsPackagesByPackageId(String packageId, String contentType
...
-            .thenCompose(o -> {
-              CompletableFuture<PackageByIdData> future = context.getPackagesService().retrievePackage(parsedPackageId);
-              return handleDeletedPackage(future, parsedPackageId, context);
-            })
-            .thenApply(packageById -> new PackageResult(packageById, null, null))
-            .thenCompose(packageResult -> updateAccessTypeMapping(accessType, packageResult, context))
+        .requestAction(context -> context.getPackagesService().retrievePackage(parsedPackageId)
+          .thenCompose(packageByIdData -> fetchAccessType(entity, context)
+            .thenCompose(accessType -> processUpdateRequest(entity, packageByIdData, context)
+              .thenCompose(o -> {
```
oleksandr-dekin, line 250: Please rename `o` :)

oleksandr-dekin (approved):

PR #374, opened 2021-10-12
MODKBEKBJ-600: Providers, deny updating tags with empty name

Shans-Kaluhin:
## Purpose
We should deny updating tags with empty name like "  "

## Approach
Set isNotBlank instead isNotEmpty

## Learning
[MODKBEKBJ-600](https://issues.folio.org/browse/MODKBEKBJ-600)
[Differences between blank end empty](https://stackoverflow.com/questions/23419087/stringutils-isblank-vs-string-isempty)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,9 @@
+## v3.10.0 xxxx-xx-xx
+* MODKBEKBJ-600 Providers, deny updating tags with empty name
+
 ## v3.9.0 2021-10-07
 * MODKBEKBJ-596 Packages, error message when "isCustom" not provided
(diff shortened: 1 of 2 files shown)


psmagin (commented):
Do we really need to allow such cases?

NEWS.md
```diff
@@ -1,6 +1,7 @@
 ## v3.9.0 2021-10-07
 * MODKBEKBJ-596 Packages, error message when "isCustom" not provided
 * MODKBEKBJ-597 Packages, deny creating package with empty name
+* MODKBEKBJ-600 Providers, deny updating tags with empty name
```
psmagin, line 4: Move it to a new version that is not released yet.

PR #375, opened 2021-10-12
MODKBEKBJ-598: Packages, clarify error message when unselect non-custom package

Shans-Kaluhin:
## Purpose
When isCustom and isSelected are false, we need to met this five conditions:
```
allowKbToAddTitles should be false,
visibilityData.isHidden should be false,
customCoverage should be empty,
packageToken.value should be null.
```
And when one of this conditions not met, server respond with message like: "Package's 'isCustom' and 'isSelected' are 'false'".
the error message is very strict, we need to remove it and display the parameter that prevents the update

## Approach
Remove pointless message and display real condition that fall

## Learning
[MODKBEKBJ-598](https://issues.folio.org/browse/MODKBEKBJ-598)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## v3.9.0 2021-10-07
 * MODKBEKBJ-596 Packages, error message when "isCustom" not provided
 * MODKBEKBJ-597 Packages, deny creating package with empty name
+* MODKBEKBJ-598 Packages, clarify error message when unselect non-custom package 
 
 ## v3.8.0 2021-06-10
 * MODKBEKBJ-570 Filter resources in title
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -1,6 +1,7 @@
 ## v3.9.0 2021-10-07
 * MODKBEKBJ-596 Packages, error message when "isCustom" not provided
 * MODKBEKBJ-597 Packages, deny creating package with empty name
+* MODKBEKBJ-598 Packages, clarify error message when unselect non-custom package
```
psmagin, line 4: This will be in the next version. Move it to v3.10.0

PR #401, opened 2022-04-06
MODKBEKBJ-630 endpoint /eholdings/packages not support filter[custom]…

YahorSviarchkou:
## Purpose
Add support request with parameter filter[custom] = false

## Approach
Changing the validation in a Filter

## Learning
[MODKBEKBJ-630](https://issues.folio.org/browse/MODKBEKBJ-630)

### DIFF
--- a/src/main/java/org/folio/rest/util/RestConstants.java
+++ b/src/main/java/org/folio/rest/util/RestConstants.java
@@ -27,12 +27,6 @@ public final class RestConstants {
       "ebsco", "orderedthroughebsco"
     );
 
-  public static final Map<String, String> FILTER_CUSTOM_MAPPING =
-    Map.of(
-      "true", "custom",
-      "false", "notcustom"
-    );
-
   public static final List<String> SUPPORTED_PACKAGE_FILTER_TYPE_VALUES =
     List.of("all", "aggregatedfulltext", "abstractandindex", "ebook", "ejournal", "print", "unknown",
       "onlinereference", "streamingmedia", "mixedcontent");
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/rest/util/RestConstants.java
```diff
@@ -27,6 +27,12 @@
       "ebsco", "orderedthroughebsco"
     );
 
+  public static final Map<String, String> FILTER_CUSTOM_MAPPING =
+    Map.of(
+      "true", "custom",
+      "false", "notcustom"
+    );
+
```
psmagin, line 35: You will never map to "custom" and "notcustom" values. You may create a list with possible values instead map

psmagin (changes requested):
Please, also create a test for this change.

src/test/java/org/folio/rest/impl/integrationsuite/EholdingsPackagesTest.java
```diff
@@ -294,6 +294,17 @@ public void shouldReturnEmptyResponseWhenPackagesReturnedWithErrorOnSearchByAcce
...
+  public void shouldReturn400OnInvalidAndNotSupportedFilterCustomParameter() {
+    String invalidParameterForFilterCustom = getWithStatus(PACKAGES_ENDPOINT + "?filter[custom]=test", 400, STUB_TOKEN_HEADER)
+      .asString();
+    String falseParameterForFilterCustom = getWithStatus(PACKAGES_ENDPOINT + "?filter[custom]=false", 400, STUB_TOKEN_HEADER)
+      .asString();
+
+    assertTrue(invalidParameterForFilterCustom.contains("Invalid Query Parameter for filter[custom]: only 'true' is supported"));
+    assertTrue(falseParameterForFilterCustom.contains("Invalid Query Parameter for filter[custom]: only 'true' is supported"));
+  }
+
```
psmagin, line 307: Please, split up it into 2 tests with 1 request in each of them..

PR #402, opened 2022-04-07
MODKBEKBJ-622 Remove parameters from POST request of '/eholdings/kb-credentials

DieHardJava:
## Purpose
Remove user-related fields from POST api call because user data is queried from mod-users, not from database.

## Approach
remove "firstName", "lastName", "patronGroup", "userName" from api;
update repository interaction (save only kb_id to user_id relation).
create new request body for POST request of '/eholdings/kb-credentials


ramls/types/assignedUsers/assignedUserPutRequest.json
```diff
@@ -1,15 +1,15 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "title": "PUT Request To Assigned Users Collection Schema",
```
psmagin, line 3: I think we can remove PUT endpoint for assigned users. It will never be used

src/main/java/org/folio/rest/converter/assignedusers/AssignedUserCollectionItemConverter.java
```diff
@@ -37,19 +38,24 @@ public AssignedUser convert(DbAssignedUser source) {
   }
 
   @Component
-  public static class ToDb implements Converter<AssignedUser, DbAssignedUser> {
+  public static class toAssignedUserId implements Converter<DbAssignedUser, AssignedUserId> {
```
psmagin, line 41:
```suggestion
  public static class ToAssignedUserId implements Converter<DbAssignedUser, AssignedUserId> {
```

psmagin (changes requested):

psmagin (approved):

Igor-Gorchakov (approved):

Shans-Kaluhin (approved):

PR #403, opened 2022-04-11
MODKBEKBJ-621 Rework GET request for '/kb-credentials/{kb-credentials-id}/users'

DieHardJava:
## Purpose
Rework GET request for '/kb-credentials/{kb-credentials-id}/users'

## Approach
get assigned users from db (only user ids by kb_credentials_id);
query users info from `mod-users` for assigned users using "/users" api and cql (Check the max possible amount of users we can query this way and implement batches so if for example we have 200 assigned users in db and the limit is 50 then we should execute 4 queries with 50 users each and compose results then);
query groups info using "/groups" api of  `mod-users`;
aggregate received info and pass as response.


src/main/java/org/folio/service/assignedusers/AssignedUsersServiceImpl.java
```diff
@@ -40,12 +48,53 @@
...
+      .thenApply(dbAssignedUsers1 -> dbAssignedUsers1.stream()
+        .map(DbAssignedUser::getId)
+        .collect(Collectors.toList()));
+
+    var users = new ArrayList<CompletableFuture<User>>();
+
+    ids.thenApply(idList -> idList.stream().map(id -> {
+      okapiHeaders.put(XOkapiHeaders.USER_ID, String.valueOf(id));
+      return users.add(usersLookUpService.lookUpUser(new OkapiParams(okapiHeaders)));
+    }));
```
psmagin, line 70: You should request a collection of users in one request. For this purpose use CQL query like:
`query=id=<id1> or id=<id2> or ....`
Please remember that url and response body have limitations, so request users in batches.

src/main/java/org/folio/service/assignedusers/AssignedUsersServiceImpl.java
```diff
@@ -40,12 +48,53 @@
...
+      okapiHeaders.put(XOkapiHeaders.USER_ID, id);
+      return groups.thenApply(g -> {
+        try {
+          return g.add(usersLookUpService.lookUpGroup(new OkapiParams(okapiHeaders)).get());
+        } catch (InterruptedException | ExecutionException e) {
+          e.printStackTrace();
+          return null;
+        }
+      });
+    }));
```
psmagin, line 87: Same for groups, use cql query

src/main/java/org/folio/service/assignedusers/AssignedUsersServiceImpl.java
```diff
@@ -40,12 +48,53 @@
...
+    var groupIds = users.stream()
+      .map(user -> user.thenApply(User::getPatronGroup)).collect(Collectors.toList());
+
+    var groups = new CompletableFuture<ArrayList<Group>>();
+
+    groupIds.forEach(groupId -> groupId.thenApply(id -> {
+      okapiHeaders.put(XOkapiHeaders.USER_ID, id);
+      return groups.thenApply(g -> {
+        try {
+          return g.add(usersLookUpService.lookUpGroup(new OkapiParams(okapiHeaders)).get());
```
psmagin, line 81: You shouldn't finalize futures. Just combine futures to get the expected result.

src/main/java/org/folio/service/assignedusers/AssignedUsersServiceImpl.java
```diff
@@ -40,12 +48,53 @@
...
+      });
+    }));
+
+    var assignedUserCollection = dbAssignedUsers
       .thenApply(collectionConverter::convert);
+    var c = new AtomicInteger();
+    assignedUserCollection.thenCombine(groups,
+      (assignedUserCollection1, groupsList) -> assignedUserCollection1.getData()
+        .stream().peek(assignedUser ->
+          assignedUser.getAttributes().setPatronGroup(groupsList.get(c.incrementAndGet()).getGroup())));
```
psmagin, line 95: You can't be sure that the groups' collection has the same side or in the same order. Use group id to identify which group name to use.

src/main/java/org/folio/service/users/UsersLookUpService.java
```diff
@@ -112,4 +135,17 @@ private User mapUser(JsonObject user) {
...
     return builder.build();
   }
+
+  private Group mapGroup(JsonObject group) {
+    Group.GroupBuilder builder = Group.builder();
+    if (group.containsKey("group")) {
+      builder.id(group.getString("id"));
+      builder.group(group.getString("group"));
+      builder.desc(group.getString("desc"));
+      builder.expirationOffsetInDays(group.getInteger("expirationOffsetInDays"));
```
psmagin, line 145: Why do you need these fields? If you don't use them, just not map them.

psmagin (changes requested):

psmagin (approved):

viacheslavkol (approved):

PR #444, opened 2023-01-27
MODKBEKBJ-713 Load holdings: process failing after exceeding status request retries

Shans-Kaluhin:
## Purpose
If job was not started, HoldingsIQ will response with empty status body

## Approach
Return NONE status if status is null. So the job can be started

### Learning
https://issues.folio.org/browse/MODKBEKBJ-713

### DIFF
--- a/src/main/java/org/folio/service/holdings/AbstractLoadServiceFacade.java
+++ b/src/main/java/org/folio/service/holdings/AbstractLoadServiceFacade.java
@@ -6,6 +6,7 @@ import static org.folio.repository.holdings.HoldingsServiceMessagesFactory.getSn
 import static org.folio.repository.holdings.HoldingsServiceMessagesFactory.getSnapshotFailedMessage;
 import static org.folio.repository.holdings.LoadStatus.COMPLETED;
 import static org.folio.repository.holdings.LoadStatus.IN_PROGRESS;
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/service/holdings/AbstractLoadServiceFacade.java
```diff
@@ -98,14 +99,14 @@ public void loadHoldings(LoadHoldingsMessage message) {
   protected abstract CompletableFuture<Void> loadHoldings(LoadHoldingsMessage message, LoadService loadingService);
 
   private CompletableFuture<HoldingsStatus> populateHoldingsIfNecessary(LoadService loadingService) {
-    return waitForStatus(statusRetryCount, loadingService).thenCompose(loadStatus -> {
+    return getLastLoadingStatus(loadingService).thenCompose(loadStatus -> {
```
psmagin, line 102: So, module does not wait for completed status now? How then it work? When the loading is starting?

Shans-Kaluhin: No, It's not waiting for status before starting the job.

psmagin (approved):

hasanepam (approved):

